PR #8, opened 2016-06-20
Fix theme select options.

kubov:


alexandrudima:
This change would actually cause the theme switcher to stop working as the value of the option is used.

PR #85, opened 2016-08-01
style(monarch.css): Add monospace default for code

ZombieHippie:
Not every OS has `Consolas` as an installed font, for those browsers, default to `monospace`.

Previously:
image


alexandrudima:
Nice! Thank you!

PR #242, opened 2016-10-19
Add YAML samples

Hirse:
Follow up from https://github.com/Microsoft/monaco-languages/pull/12.

I am not sure, if it is correct to include the generated file. I have updated and included it because it existed before.


alexandrudima:
:+1: Thank you! -- the file is "duplicated" to allow for loading it on the `file://` protocol (without using xhr) from the test pages.

PR #315, opened 2017-01-13
Fix default folding value

leMaik:
It seems to default to false, but the docs say it defaults to true.

### DIFF
--- a/website/playground/monaco.d.ts.txt
+++ b/website/playground/monaco.d.ts.txt
@@ -1290,7 +1290,7 @@ declare module monaco.editor {
         codeLens?: boolean;
         /**
          * Enable code folding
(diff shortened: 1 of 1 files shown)


alexandrudima:
Thanks, this is already fixed on master.

https://github.com/Microsoft/vscode/blob/master/src/vs/editor/common/editorCommon.ts#L449
```
	/**
	 * Enable code folding
	 * Defaults to true in vscode and to false in monaco-editor.
	 */
	folding?: boolean;
```

leMaik:
@alexandrudima Oh, didn't check it there. When will it land here?

alexandrudima:
I plan to do it in the following days.

PR #497, opened 2017-07-05
Update custom-languages playground sample to use SnippetString

microsoftsam:
Since the 0.9.0 release, the monaco-editor no longer supports the build-in snippet syntax. 

This PR updates the playground samples to use SnippetString and updates the syntax of the snippets to use the TextMate syntax described here: https://code.visualstudio.com/docs/editor/userdefinedsnippets#_snippet-syntax


alexandrudima:
Thank you! :heart:

PR #726, opened 2018-02-19
Add linebreak for if

MSeifert04:
This just adds a line break in an playground example that makes the code more readable (in my opinion).


alexandrudima:
:+1:

Issue #191, opened 2016-09-15
Set schema of JSON content without $schema property


Hirse:
In [monaco-json](https://github.com/Microsoft/monaco-json), it is possible to[ assign json schemas](https://github.com/Microsoft/monaco-json/blob/master/src/monaco.d.ts#L27) (for validation and code hints) to json files (file-names), however when running monaco stand alone in a browser, I don't have any filenames.

How can I select a schema to use for a json editor instance?

Using the `$schema` property works in theory, but gives problems if the schema doesn't allow this property in the root scope. See the following example modified from the [Monaco Editor Playground JSON example](https://microsoft.github.io/monaco-editor/playground.html#extending-language-services-configure-json-defaults):
```js
monaco.languages.json.jsonDefaults.setDiagnosticsOptions({
    validate: true,
	schemas: [{
        uri: "http://myserver/foo-schema.json",
        schema: {
            properties: {
                p1: {}
            },
            additionalProperties: false
        }
    }]
});

var jsonCode = [
	'{',
	'    "$schema": "http://myserver/foo-schema.json"',
	"}"
].join('\n');

monaco.editor.create(document.getElementById("container"), {
	value: jsonCode,
	language: "json"
});
```

![2016-09-15_16-12-08](https://cloud.githubusercontent.com/assets/2564094/18553042/4ee83a8a-7b5f-11e6-996d-5a9efbcb28dc.png)

--
monaco-editor npm version: 0.6.1
Browser: All
OS: All

aeschli:
Here's a example on how to do it. In the file match you can also use patterns like `*.json`, or `foo*.json`.


```typescript

monaco.languages.json.jsonDefaults.setDiagnosticsOptions({
	schemas: [{
      fileMatch: ['foo.json'],
      schema: {
            type: "object",
            properties: {
                p1: {
                    enum: [ "v1", "v2"]
                }
            }
        }
    }]
});


var jsonCode = [
	'{',
	"}"
].join('\n');

var model = monaco.editor.createModel(jsonCode, 'json', 'internal://server/foo.json');

monaco.editor.create(document.getElementById("container"), {
	model: model
});
```

Hirse:
@aeschli Thanks for the explanation. I couldn't find the information about the name as identifier before.

apalsapure:
@aeschli Thanks for the sample. 
[...]

PR #731, opened 2018-02-21
Fix `Configure JSON defaults` sample

apalm:
The [Configure JSON defaults sample](https://microsoft.github.io/monaco-editor/playground.html#extending-language-services-configure-json-defaults) in the playground is currently broken. Validation errors never show up even when the JSON doesn't conform to the schema.

screen shot 2018-02-20 at 20 15 52

Here's what I had to change to get it to work:
* Set `validate` to `true` in diagnostic options.
* Create a model; for each schema, set `fileMatch` to a value that matches the `uri` used when creating the model. Reference: https://github.com/Microsoft/monaco-editor/issues/191#issuecomment-248034168

I also changed the initial JSON so that the validation errors show up without editing the JSON.

### Test Plan
Ran `gulp simpleserver`. The validation errors now show up as expected.

screen shot 2018-02-20 at 20 16 46

### DIFF
--- a/test/playground.generated/extending-language-services-configure-json-defaults.html
+++ b/test/playground.generated/extending-language-services-configure-json-defaults.html
@@ -38,9 +38,13 @@ loadEditor(function() {
 
 // Configures two JSON schemas, with references.
 
+var id = "foo.json";
+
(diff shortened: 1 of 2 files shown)


alexandrudima:
Thank you! :+1:

PR #740, opened 2018-03-03
Fix line number API usage

nikeee:
This PR fixes the samples that are using the deprecated API of lineNumbers (boolean). They now use `"on"`, `"off"` and `"relative"`.


rebornix:
@nikeee thanks for updating the API usage ;)

PR #878, opened 2018-05-15
Add folding provider sample to the playground

rcjsuen:
A new `monaco.languages.registerFoldingRangeProvider` API was added in 0.13.0 to allow people to create custom folding regions. I've created a small example that demonstrates how to use this API for the Monaco Playground.


alexandrudima:
:+1: Thank you!

PR #916, opened 2018-06-13
Fix hash comment in xdot sample

nikeee:
The xdot comment using hash does only include one following character. It should be like `//`, spanning to the end of the line.


alexandrudima:
:+1:

PR #957, opened 2018-07-15
fix small error in integration docs

cancerberoSgx:
A small mistake fixed. `baseUrl` should be the same as in the option 2 (read below in the same document) . If not it won't work and can give unnecessary pain to newcomers. Good work :) keep it up!


alexandrudima:
:+1: Thank you!

PR #962, opened 2018-07-18
Fix playground sample option

haeguri:
The `nativeContextMenu` property did not work, so I changed it to the` contextmenu` property.


test/playground.generated/interacting-with-the-editor-listening-to-mouse-events.html
```diff
@@ -60,7 +60,7 @@
 	value: jsCode,
 	language: "javascript",
 	glyphMargin: true,
-	nativeContextMenu: false
+	contextmenu: false
```
haeguri: f

alexandrudima:
Thank you! :heart:

Issue #1021, opened 2018-08-22
registerHoverProvider typings make "no hover" impossible to express


fabsenet:
monaco-editor version: [playground](https://microsoft.github.io/monaco-editor/playground.html#extending-language-services-hover-provider-example)
Browser: firefox
OS: Windows 7
Steps or JS usage snippet reproducing the issue:
- open playground
- replace javascript with

```js
monaco.languages.register({ id: 'mySpecialLanguage' });

monaco.languages.registerHoverProvider('mySpecialLanguage', {
    provideHover: function (model, position) {
        if (position.lineNumber !== 3) return; 
        // ^^^^ this provider only works for line 3 but typescript fails here!!
        return {
            range: new monaco.Range(3, 1, 3, model.getLineMaxColumn(3)),
            contents: [
                { value: 'title' },
                { value: 'some text' }
            ]
        }
    }
}
);
monaco.editor.create(document.getElementById("container"), {
    value: 'abc\ndef\nHover ONLY over THIS text',
    language: 'mySpecialLanguage'
});

```
Result: Hovering works over line 3, does not do anything when hovering over line 1 or 2.

Simply returning or returning undefined or null or an empty object is not supported by the typings in editor.api.d.ts
```ts
 provideHover(model: editor.ITextModel, position: Position, token: CancellationToken): Hover | Thenable<Hover>;
```

Can you change the expected return type to: `Hover | Thenable<Hover> | null` ?
I cannot build a hover provider in typescript which does not provide hovers for everything.

fabsenet:
Or better, the return type should be

`Hover | undefined | null | Thenable<Hover | undefined | null>`

I checked the using code and verified it with faking the result and it works so far.

"Faked" provider as a workaround:

```typescript
    monaco.languages.registerHoverProvider("myLang", {
      provideHover: function(model, position) {
        let ret: monaco.languages.Hover;
        if (position.lineNumber == 3) {
          ret = {
            range: new monaco.Range(1,1,1,4),
            contents: [{ value: "a" }, { value: "b" }],
          };
        }
        return ret!; // the ! tells the typescript compiler to shut up
      },
    });
```

alexandrudima:
@fabsenet Can you make a PR please?

fabsenet:
this issue is not fixed.

alexandrudima:
@fabsenet Why is it not fixed ?

fabsenet:
[...]

PR #1030, opened 2018-08-27
Aligned provideHover signature to actual implementation

fabsenet:
fixes https://github.com/Microsoft/monaco-editor/issues/1021

I checked the actual working in real code for this change but I have no idea how the build and/or tests in this project work so this change is only 85% safe from my point of view, sorry.

### DIFF
--- a/monaco.d.ts
+++ b/monaco.d.ts
@@ -4723,7 +4723,7 @@ declare namespace monaco.languages {
          * position will be merged by the editor. A hover can have a range which defaults
          * to the word range at the position when omitted.
          */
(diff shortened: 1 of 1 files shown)


alexandrudima:
@fabsenet Sorry that file is a snapshot (checked in this repository such that the website gets good intellisense) of a generated file (monaco.d.ts is generated from the editor sources). Can you pls make the change at https://github.com/Microsoft/vscode/blob/620d2e95c27eaa4990ab60f0e2110bd28573739e/src/vs/editor/common/modes.ts#L247

fabsenet:
followup PR https://github.com/Microsoft/vscode/pull/57270

PR #1041, opened 2018-08-30
Point to the 'release' folder for 'esm' and 'typings'.

DanielRosenwasser:


alexandrudima:
@DanielRosenwasser That would be incorrect. The version coming in from `npm` does not have a `release` folder. https://registry.npmjs.org/monaco-editor/-/monaco-editor-0.14.3.tgz

PR #1094, opened 2018-10-08
Added an small example for new developers

KayDee:
New developers might need a little simplification. I know i did.


alexandrudima:
Thank you, but we will not take this PR at this time.

KayDee:
If its ok to ask, will this PR be considered for Hacktoberfest (Microsoft special) ?

KayDee:
@berndverst. Could you confirm if i filled that form for the hacktoberfest ? I don't seem to remember and the form is closed now.

Also, could you answer the above question as well.

PR #1108, opened 2018-10-12
Add linebreak in website/index.html

hyundw84:
This just adds a line break in website/index.html that makes the code more readable (in my opinion).


alexandrudima:
Thank you, we will not take this PR at this time.

PR #1113, opened 2018-10-13
Remove commented code

KadoBOT:
Commented code should be avoided. There are more downsides then upsides on keeping it. If this piece of code is needed in the future, `git diff` is a better option.


alexandrudima:
:+1: Thank you.

PR #1125, opened 2018-10-18
Contribution_Fr

arpitv424:


alexandrudima:
Thank you, but we have no plans to provide developer documentation in other languages than English.

PR #1126, opened 2018-10-18
Add selectable issue templates

myzeprog:
This PR adds selectable issue templates to the repository.

This repository already has a prefilled text when creating an issues. It tells users that they shouldn't make issues for support questions, but these types of issues are still made regularly. Adding selectable issue templates might be more effective at preventing erroneous issues from being created.


alexandrudima:
Thank you, but for now we want to keep things simple.

PR #1144, opened 2018-10-25
Add allowNonTsExtensions to CompilerOptions

trevorade:
I don't know why `allowNonTsExtensions` is missing from `CompilerOptions`.  This bit me recently as you _have_ to set `allowNonTsExtensions: true` when using in-memory models.  When running in compiled mode, this property got renamed as it was missing from `monaco.d.ts`

Note: I didn't add the newline at the end of the file.  I edited the file directly in github and its editor appears to have added it as part of saving my change.

### DIFF
--- a/monaco.d.ts
+++ b/monaco.d.ts
@@ -5451,6 +5451,7 @@ declare namespace monaco.languages.typescript {
     type CompilerOptionsValue = string | number | boolean | (string | number)[] | string[] | MapLike<string[]> | null | undefined;
     interface CompilerOptions {
         allowJs?: boolean;
+        allowNonTsExtensions?: boolean;
         allowSyntheticDefaultImports?: boolean;
         allowUnreachableCode?: boolean;
         allowUnusedLabels?: boolean;


trevorade:
I'm unclear if `monaco.d.ts` is generated or what. If it _is_ generated, I don't know where it's getting the `CompilerOptions` interface properties from.

trevorade:
+@alexandrudima

Are there any next steps I'm missing here?

alexandrudima:
The source of the file is in the monaco-typescript repository. But I don't understand where this flag (`allowNonTsExtensions`) is coming from. The interface in [monaco-typescript](https://github.com/Microsoft/monaco-typescript/blob/609b93eb011be5b9840d988249d4fbad012130ed/src/monaco.d.ts#L46:L123) is taken from the official [TypeScript API](https://github.com/Microsoft/monaco-typescript/blob/609b93eb011be5b9840d988249d4fbad012130ed/src/lib/typescriptServices.d.ts#L2416).

Perhaps you want to create this PR against TypeScript, then with the next update we will pick a new compiler version in monaco-typescript, and finally that will eventually make it into monaco-editor.

PR #1147, opened 2018-10-26
Refactoring website/playground to ES6 

Archaeologist03:
Refactoring vars to let and consts and string concatenation to template literal.


alexandrudima:
Thanks, but we don't want to force our visitors to have ES6 browsers yet.

PR #1268, opened 2019-01-10
Added documentation for using parcel.

AaronGillBraun:
The sample using parcel ([browser-esm-parcel](https://github.com/Microsoft/monaco-editor-samples/tree/master/browser-esm-parcel)) does not work due the how parcel handles workers. Added documentation showing how parcel can still be used by bundling workers separately.


alexandrudima:
:+1: Thank you

PR #1329, opened 2019-02-26
Fix JavaScript RegExp range closing bracket

jbsulli:
The Monarch tokenizer for JavaScript has a bug where the closing bracket for the RegExp range refers to a `brackets` attribute that does not exist. This bug causes the closing bracket to not be colored using the regexp style.

### DIFF
--- a/website/monarch.html
+++ b/website/monarch.html
@@ -772,7 +772,7 @@ return {
 			[/\^/, 'regexp.invalid'],
 			[/@regexpesc/, 'regexp.escape'],
 			[/[^\]]/, 'regexp'],
-			[/\]/, '@brackets.regexp.escape.control', '@pop'],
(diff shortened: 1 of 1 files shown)


alexandrudima:
:+1: Thank you!

jbsulli:
@alexandrudima Happy to help! My change was just for the editable example. I couldn't find where the actual JavaScript Monarch language file is. It seems to have the same issue. Is it in a private repo?

alexandrudima:
@jbsulli I've applied your change also to the `monaco-languages` repository, where the grammars we ship live -- https://github.com/Microsoft/monaco-languages/commit/ba442264340ddd4f5cac5421f3a2eca343d828c1

jbsulli:
Awesome! Thank you so much!

PR #1337, opened 2019-03-01
Fix lineDecoration example css error

datou0412:
'left' style will not work in linesDecorationsClassName, but 'margin-left' do


alexandrudima:
:+1: Thank you!

PR #1339, opened 2019-03-02
fix worker paths in parcel

cancerberoSgx:
I just tried this instructions and I found this small detail for they to work perfectly

### DIFF
--- a/docs/integrate-esm.md
+++ b/docs/integrate-esm.md
@@ -138,18 +138,18 @@ import * as monaco from 'monaco-editor';
 self.MonacoEnvironment = {
   getWorkerUrl: function(moduleId, label) {
     if (label === 'json') {
-      return './json.worker.bundle.js';
+      return './json.worker.js';
     }
     if (label === 'css') {
-      return './css.worker.bundle.js';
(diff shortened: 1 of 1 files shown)


jbsulli:
Interesting. I'm using a Windows machine so the shell script wasn't an option. I used the following with Parcel and the script isn't needed:

```javascript
self.MonacoEnvironment = {
  getWorker(_moduleId: string, label: string) {
    switch (label) {
      case "json":
        return new Worker("../node_modules/monaco-editor/esm/vs/language/json/json.worker.js");
      case "css":
        return new Worker("../node_modules/monaco-editor/esm/vs/language/css/css.worker.js");
      case "html":
        return new Worker("../node_modules/monaco-editor/esm/vs/language/html/html.worker.js");
      case "typescript":
      case "javascript":
        return new Worker("../node_modules/monaco-editor/esm/vs/language/typescript/ts.worker.js");
      default:
        return new Worker("../node_modules/monaco-editor/esm/vs/editor/editor.worker.js");
    }
  }
};
```

alexandrudima:
:+1: Thank you!

PR #1354, opened 2019-03-09
changed nav-item:hover background color

snturk:
nav-item:hover background color has changed with a color which is more pronounced/noticeable and compatible with website's theme.


alexandrudima:
I'm not even sure what this change impacts. Could you provide before/after screenshots?

snturk:
I just changed the hover color on navigation bar's items to noticeable one. But now I look the code from website and can't see same CSS codes about hover. So I can't send you screenshots.

PR #1388, opened 2019-03-23
CHANGELOG: Fix year for releases made in 2019

jonas:


Hongarc:
@aeschli This pull request includes Microsoft@646bf45 and Microsoft@37035ef

Why you don't merge instead of use that commits.

PR #1445, opened 2019-05-17
make range in CompletionItem optional and fix symbol links in comments

vemoo:
The documentation of [range](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.languages.completionitem.html#range) says it has a default value, I've also tested it and it works.

I also realised that the links to symbols y comments where not working, so I've fixed them.

### DIFF
--- a/monaco.d.ts
+++ b/monaco.d.ts
@@ -1583,7 +1583,7 @@ declare namespace monaco.editor {
         /**
          * Converts the position to a zero-based offset.
          *
-         * The position will be [adjusted](#ITextModel.validatePosition).
+         * The position will be [adjusted]{@link ITextModel.validatePosition}.
          *
          * @param position A position.
          * @return A valid zero-based offset.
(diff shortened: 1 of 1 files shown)


monaco.d.ts
```diff
@@ -4807,13 +4807,13 @@ declare namespace monaco.languages {
         /**
          * A range of text that should be replaced by this completion item.
          *
-         * Defaults to a range from the start of the [current word](#TextDocument.getWordRangeAtPosition) to the
+         * Defaults to a range from the start of the [current word]{@link ITextModel.getWordAtPosition} to the
```
poffdeluxe: I believe it actually uses [`getWordUntilPosition`](https://github.com/microsoft/vscode/blob/master/src/vs/editor/contrib/suggest/suggest.ts#L125)

(Thanks for making this PR btw.. I came here to do the exact same thing)

alexandrudima:
I am sorry, `monaco.d.ts` is a generated file from the sources at https://github.com/microsoft/vscode

PR #1450, opened 2019-05-21
Fix C# sample code

DaMightyZombie:
String literal does not need to be @-escaped.
Alternatively, one could remove one of the backslashes in the line.


alexandrudima:
The `@` is used to test/show off the colorizer.

PR #1543, opened 2019-08-09
Add Sophia ML example

mradkov:
Add Sophia ML example.

This is associated with https://github.com/microsoft/monaco-languages/pull/67


alexandrudima:
Thank you! :heart:

PR #1595, opened 2019-09-23
Make range optional in CompletionItem definition

FreHu:
This PR fixes the type error `Property 'range' is missing in type '{ label: string; kind: CompletionItemKind.Function; documentation: string; insertText: string; }' but required in type 'CompletionItem'.`

The range is not required. It is stated in the docstring that it 
`Defaults to a range from the start of the [current word](#TextDocument.getWordRangeAtPosition) to the current position.`

It is not included in the example code either:
https://microsoft.github.io/monaco-editor/playground.html#extending-language-services-completion-provider-example


alexandrudima:
@FreHu `monaco.d.ts` is generated from the VS Code sources.

The "truth" is [here](https://github.com/microsoft/vscode/blob/00983bcdf0bd747967c6bea520ba695c4152e9c3/src/vs/editor/common/modes.ts#L455)

PR #1688, opened 2019-11-28
Make token optional in ITokenThemeRule interface

remi-deltombe:
Theme options should allow rules without token.

Example : 
```
    /// https://microsoft.github.io/monaco-editor/playground.html#customizing-the-appearence-exposed-colors
    monaco.editor.defineTheme('myTheme', {
        rules: [{ background: 'EDF9FA' }],
    });
```

### DIFF
--- a/monaco.d.ts
+++ b/monaco.d.ts
@@ -963,7 +963,7 @@ declare namespace monaco.editor {
     };
 
     export interface ITokenThemeRule {
-        token: string;
+        token?: string;
         foreground?: string;
         background?: string;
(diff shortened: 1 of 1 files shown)


alexdima:
The `monaco.d.ts` file is a generated file. It is generated from VS Code sources. In this case, the truth of the interface is at https://github.com/microsoft/vscode/blob/da765520e8092fd39384744ae90147cd1bbcdefa/src/vs/editor/common/modes/supports/tokenization.ts#L9-L14

PR #1756, opened 2020-01-08
Playground: Add keyboard shortcut to run playground code

brijeshb42:
<kbd>CTRL/CMD + Enter</kbd> will run the current code in the playground editor.


alexdima:
Thank you! :heart:

PR #1785, opened 2020-01-27
Validate typings can be imported correctly in simple ts project

rebornix:
Ensure there is no error when using monaco-editor-core in a ts project


alexdima:
:+1: Nice !!

PR #1922, opened 2020-04-15
Playground: add sample for adding external lib and exposing it globally

StasZelenskiy:


mike-lischke (commented):
Very helpful example, please merge this patch!

alexdima:
This sample uses exactly the same monaco editor APIs as https://microsoft.github.io/monaco-editor/playground.html#extending-language-services-configure-javascript-defaults , so I believe it is redundant. 

I agree that this sample shows how it is possible to add two different `.d.ts` files and then import from one, and export to the global list from the other, but that is very specific to the TypeScript language. There are very many patterns in the TypeScript language that can be used in `.d.ts` files and the scope of this playground is limited to show editor API. In this case, the used editor API is equivalent to the editor API used in https://microsoft.github.io/monaco-editor/playground.html#extending-language-services-configure-javascript-defaults .

So what I am saying is that while this sample is very useful, it is redundant from an editor API point of view, as the used editor API is already covered in another sample.

mike-lischke:
OK, I mostly agree, even though I find the typings addition aspect pretty hidden behind this configure defaults example. I'd prefer a dedicated demo, **particulary** because this is a pretty specific thing. Just my 2 cents.

PR #1961, opened 2020-05-10
feat: fix and migrate smoke tests to Playwright

mxschmitt:
- Migrated and fixed the smoke tests over to Playwright and execute them on webkit/firefox/chromium as discussed with @rebornix
- Adjusted the GitHub Actions, so they run also on a Pull Requests


rebornix:
Thanks for the good work!

PR #1964, opened 2020-05-12
ci: fixed smoke tests by increasing timeout

mxschmitt:
Currently CI is still failing on GitHub Actions.
My guess is that Firefox takes sometimes longer than 2 seconds which is the default timeout, thats why Mocha is failing because it timeouts.
Should be a quick fix for now. Since we only want to check the basic functionality, it should be fine, otherwise I would recommend to e.g. migrate to Jest.


mxschmitt:
cc @rebornix

Thank you!

mxschmitt:
cc @rebornix

Thank you!

alexdima:
Thank you!

Issue #1833, opened 2020-02-14
Semantic highlighting does not appear to work


rcjsuen:
1. I expect the first four characters to be highlighted but nothing happens.
1. There is no documentation really for any of the functions but `releaseDocumentSemanticTokens` in particular is not clear to me what I should be doing there.
```JavaScript
monaco.languages.register({
    id: 'new'
});
monaco.languages.registerDocumentSemanticTokensProvider('new', {
    getLegend: () => {
        return {
            tokenModifiers: [],
            tokenTypes: [
                "keyword"
            ]
        }
    },

    provideDocumentSemanticTokens: () => {
        return {
            data: [
                0, 0, 4, 0, 0
            ]
        }
    },

    releaseDocumentSemanticTokens: () => {
    }
});

monaco.editor.create(document.getElementById("container"), {
    value: "12345678",
    language: "new"
});
```

alexdima:
The standalone APIs are all hooked up, but I believe there is no theming implemented for the standalone editor. So those semantic tokens get looked up, no theming returns, so the semantic tokens are dropped.

https://github.com/microsoft/vscode/blob/37f2127173cf8d3185ec0ea27a7643fdcddbb07b/src/vs/editor/standalone/browser/standaloneThemeServiceImpl.ts#L134-L136

rcjsuen:
@alexdima @aeschli Is it possible for me to workaround this (by calling internal APIs or whatever) or is it all too tightly coupled that I should just wait for a bug fix release?

Thank you for your information.

alexdima:
@rcjsuen You could try to do this by providing a custom theme service, (with the first created editor) e.g.

```ts
monaco.editor.create(element, opts, { themeService: yourCustomImplementation });
```

You would then need to implement `IThemeService `:
```ts
export interface IThemeService {
	_serviceBrand: undefined;

	getTheme(): ITheme;

	readonly onThemeChange: Event<ITheme>;

	getIconTheme(): IIconTheme;

	readonly onIconThemeChange: Event<IIconTheme>;

}
```

rcjsuen:
@alexdima I tried a simpler approach but nothing seems to be working. Since `provideDocumentSemanticTokens` is not being printed to the console I am of the belief that Monaco is not even polling the provider for tokens. Nothing is being printed from my replaced `getTokenStyleMetadata` function either.
```JavaScript
monaco.languages.register({
    id: 'new'
});
[...]

PR #2103, opened 2020-09-05
Semantic tokens provider example

KapitanOczywisty:
This is simplified version of [VSCode Semantic tokens sample](https://github.com/microsoft/vscode-extension-samples/tree/master/semantic-tokens-sample) adapted for monaco editor.

### DIFF
--- a/website/playground/new-samples/extending-language-services/semantic-tokens-provider-example/sample.js
+++ b/website/playground/new-samples/extending-language-services/semantic-tokens-provider-example/sample.js
@@ -1,0 +1,162 @@
+/** @type {monaco.languages.SemanticTokensLegend} */
+const legend = {
+    tokenTypes: [
+        'comment', 'string', 'keyword', 'number', 'regexp', 'operator', 'namespace',
+        'type', 'struct', 'class', 'interface', 'enum', 'typeParameter', 'function',
+        'member', 'macro', 'variable', 'parameter', 'property', 'label'
+    ],
+    tokenModifiers: [
+        'declaration', 'documentation', 'readonly', 'static', 'abstract', 'deprecated',
+        'modification', 'async'
+    ]
+};
+
+/** @type {(type: string)=>number} */
+function getType(type) {
+    return legend.tokenTypes.indexOf(type);
+}
+
+/** @type {(modifier: string[]|string|null)=>number} */
+function getModifier(modifiers) {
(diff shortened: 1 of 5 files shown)


website/playground/new-samples/extending-language-services/semantic-tokens-provider-example/sample.js
```diff
@@ -0,0 +1,162 @@
...
+An error case:
+    [notInLegend]`,
+    language: "plaintext",
+    theme: 'myCustomTheme',
+    // semantic tokens provider is disabled by default
+    'semanticHighlighting.enabled': true
+});
+
+// currently there isn't builtin token handling
+editor._themeService._knownThemes.forEach(function (theme) {
```
rcjsuen, line 141: I'm not sure if the samples should be suggesting to people to use internals. In any case, this code should at least reference #1833 in a comment so that it gets more publicity and hopefully gets fixed.

KapitanOczywisty: Maybe this is a bit ify, but if there is not other way... People are searching for this and there isn't complete example to try and play with.

Link could be useful, but probably comment with detailed explanation would be better.

rcjsuen: > Link could be useful, but probably comment with detailed explanation would be better.

Well, a link is better than no link.

KapitanOczywisty:
@alexdima since IE11 is no longer supported, can we use ES6 in playground examples?

alexdima:
Thank you!

KapitanOczywisty:
https://github.com/microsoft/vscode/commit/6f5948138c23cf857ce9609e075fcbc153219886 There is nothing more permanent than a temporary solution ;)

rcjsuen:
@alexdima When will that change go live? The example on the website doesn't show me any colours. I'm using Firefox on Windows 10.

KapitanOczywisty:
@rcjsuen From what I can see he pushed `gh-pages`, but forgot to actually build website. @alexdima ? 🤔

rcjsuen:
I will note that having a working example on the website is meaningless without an updated library pushed to npm. 🤔

alexdima:
Yes, the plan was to publish a new editor version which would have made the sample to work correctly. But I found #2140 right before publishing so I decided to postpone releasing until that is fixed.

PR #2112, opened 2020-09-08
Update example: allow peek definition

KapitanOczywisty:
This is basically implementation of feature added in https://github.com/microsoft/vscode/pull/85129


alexdima:
Thank you!

Issue #2111, opened 2020-09-08
Monaco TypeScript master broken in amd build?


orta:
**monaco-editor version:** 2a99615
**Browser:** N/A
**OS:** N/A
**Playground code that reproduces the issue:** I don't think its set up to run on commits

The TS site builds a nightly version of both Monaco Editor and Monaco TS [here](https://github.com/microsoft/TypeScript-Make-Monaco-Builds/)

The build yesterday broke at runtime in what I think is monaco-typescript code:

- Two days ago: 🟢   https://www.typescriptlang.org/play?ts=4.1.0-dev.20200906#code/MYewdgzgLgBAFgUwDZJDAvDARAdxAJyQBMsAoIA
- One day ago: 🟢  https://www.typescriptlang.org/play?ts=4.1.0-dev.20200907#code/MYewdgzgLgBAFgUwDZJDAvDARAdxAJyQBMsAoIA
- This morning: 🔴  https://www.typescriptlang.org/play?ts=4.1.0-dev.20200908#code/MYewdgzgLgBAFgUwDZJDAvDARAdxAJyQBMsAoIA

The raised issue is `n.Emitter is not a constructor` and when you go back in the trace it is requiring `monaco.contribution.ts` where it fails, [likely here](https://github.com/microsoft/monaco-typescript/blob/70d1cb17042917b13cf8e817ab41a9f8b75b2b4f/src/monaco.contribution.ts#L465-L466):

```ts
private _onDidChange = new Emitter<void>();
private _onDidExtraLibsChange = new Emitter<void>();
```

Maybe https://github.com/microsoft/monaco-typescript/commit/ee95401a407e42e1044813087ba6dc621ecc25e9 ? 

( Left notes because I'm about go to hit meetings for the day )

spahnke:
I had the same problem when I tried to put a current monaco-typescript build into the 0.20.0 editor ~10 hours ago. But I also assume this is maybe fixed by https://github.com/microsoft/monaco-editor/commit/2a996151d74d0e48102c47ea4c15836023d32f94#diff-b9e12334e9eafd8341a6107dd98510c9R199-R202 (that was ~5 hours ago, and I don't know what "morning" means to you 😄 ).

spahnke:
[...]

PR #2118, opened 2020-09-09
Adds support for both quotes in the TS side

orta:
TypeScript uses both `'` and `"` in the output, so only one of the contribution changes were being applied - re discussion in #2111 

Screen Shot 2020-09-09 at 10 29 29 AM


alexdima:
:+1: Yes, it was the minified version. Terser was converting `'` to `"` and I only really tested the dev built, since I never had problems in the past with the minified one.

PR #2186, opened 2020-10-10
Missing breaking change

luminaxster:
I'm not sure what release starting 0.21.0 caused it, not other methods being static now beyond parse() and file() in monaco.uri.
Regards and keep doing this amazing project awesomER!


alexdima:
Thank you!

PR #2250, opened 2020-12-07
Fixed indentation in C# sample

svick:
One line used spaces instead of tabs, which meant the code appeared to be incorrectly indented:


alexdima:
Thank you!

PR #2541, opened 2021-06-22
Add Bicep sample

anthony-c-martin:
Follow up from https://github.com/microsoft/monaco-languages/pull/132

Contributing this as I noticed the sample on the demo page (https://microsoft.github.io/monaco-editor/) doesn't display:
image


alexdima:
Thank you!

Issue #2507, opened 2021-05-31
Cannot set an active line number color in theme


Surm4:
According to the https://microsoft.github.io/monaco-editor/playground.html#customizing-the-appearence-exposed-colors
I can't set an active line number color in `monaco.editor.defineTheme` I can change just the non-active line numbers via `editorLineNumber.foreground`

hediet:
You can use `editorActiveLineNumber.foreground`. In general, you can use all editor colors that are available to VS Code themes.

Feel free to open a PR to update the example, we'd be very happy to merge it!

PR #2561, opened 2021-07-03
Exposed colors sample update in the playground.

Surm4:
According to the pr https://github.com/microsoft/monaco-editor/issues/2507 
I decided to update the color which was missing in the issue. 
`editorActiveLineNumber.foreground` is [deprecated](https://github.com/microsoft/vscode/blob/3718ef05565efeb27e7ee29e50ee77fc82458bb0/src/vs/editor/common/view/editorColorRegistry.ts#L28) so I've added `editorLineNumber.activeForeground` token instead.


hediet (approved):

hediet:
Thanks for this PR!

Issue #942, opened 2018-07-04
editor.addCommand returns a command id string instead of a disposable


brijeshb42:
monaco-editor version: 0.13.1
Browser: Chrome 67
OS: macOS

In this [file](https://github.com/Microsoft/vscode/blob/a6e28b0be061f27a5ea179aa675e7d8e23d6cfd3/src/vs/editor/standalone/browser/standaloneCodeEditor.ts#L195), `StandaloneCodeEditor.addCommand` returns a `commandId` instead of a disposable. Is there any specific reason for it? If yes, then how do I remove a manually added command with public api?

troy351:
@alexandrudima I got the same question.
And is there a way to dispose added command?

alexandrudima:
You could simply switch to `addAction` and also define a label. The only real difference is that the `F1` list will then show your command, which is not that bad...

bryanl:
@alexdima What you prescribe won't work for code lens out of the box unless you created some other machinery.

929853209:
I don't found a method to reject ```addCommand``` ，but I found it won't be called multiply by repeat define ```addCommand```in a react hook.  Was the ```addCommand```just like a Singleton? And I still want to know how to reject it.

alexdima:
The upcoming editor release will add `monaco.editor.registerCommand` to cover this usecase. (via https://github.com/microsoft/vscode/commit/7c937a7a383ca131c5e36811d4440878c97a1191)

PR #2616, opened 2021-08-16
Small update on the website playground

SpaceComet:
Hi. 
I added 3 small fixes. Two of them are spelling corrections and the other one is an outdated comment
We can't `dispose` `addCommand` anymore... It can create confusion .

### DIFF
--- a/website/playground/new-samples/interacting-with-the-editor/listening-to-key-events/sample.js
+++ b/website/playground/new-samples/interacting-with-the-editor/listening-to-key-events/sample.js
@@ -7,4 +7,5 @@ var myBinding = editor.addCommand(monaco.KeyCode.F9, function() {
 	alert('F9 pressed!');
 });
 
-// When cleaning up remember to call myBinding.dispose()
+// You can't dispose `addCommand`
+// If you need to dispose it you might use `addAction` or `registerCommand`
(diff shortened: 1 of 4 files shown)


website/playground/new-samples/interacting-with-the-editor/listening-to-key-events/sample.js
```diff
@@ -7,4 +7,5 @@ var myBinding = editor.addCommand(monaco.KeyCode.F9, function() {
 	alert('F9 pressed!');
 });
 
-// When cleaning up remember to call myBinding.dispose()
+// You can't dispose `addCommand`
+// If you need to dispose it you might use `addAction` or `registerCommand`
```
hediet, line 11: I don't know all about addCommand / addAction, but why would you need to dispose?

SpaceComet: I'm using monaco-editor in a NextJS app. 
I need it to be updated every time an specific hook updates. If I don't dispose it the app gets slow pretty quick.
Now I'm just using `addAction` because it can be dispose.
It seams like [it used to return a dispose](https://github.com/microsoft/monaco-editor/issues/942) then it change and someone forgot to update the playground.

SpaceComet: [Here I opened](https://github.com/suren-atoyan/monaco-react/issues/277) an issue on [monaco-react](https://github.com/suren-atoyan/monaco-react) . At thebegining I thought I could use `addCommand` because in the playground you can see the lines:

```js
// When cleaning up remember to call myBinding.dispose()
```

But, that is not true anymore. That's why I changed.

alexdima:
Thank you!

PR #2640, opened 2021-08-27
Add example for inlay hints

HKalbasi:
This PR adds example for inlay hint in Playground > Extending language service, and inlay hint colors to Exposed colors example.


alexdima:
Thank you!

PR #2644, opened 2021-08-29
Added `ariaContainerElement` to shadow dom test

crackalak:
Depends on changes made in: https://github.com/microsoft/vscode/pull/131865

Sets the container for ARIA messages to the element within the shadow dom


alexdima:
Thank you!

PR #2655, opened 2021-09-09
Make range optional for CompletionItem

JohanAhlen:


alexdima:
This file is generated from the `monaco.d.ts` file inside `microsoft/vscode`, which is also generated from https://github.com/microsoft/vscode/blob/e4c2685a278b96222a47a0af1a98b8a626b411cf/src/vs/editor/common/modes.ts#L580 . 

Changing the `monaco.d.ts` here has no effect, as the file will be overwritten with the next editor release.

PR #2674, opened 2021-09-28
Fix letter case in repo URL

silverwind:
The repo url in `package.json` is incorrectly using uppercase in the origanization name, fix that.


alexdima:
Thank you!

PR #2749, opened 2021-11-05
Update builtin functions for mysql

MasterOdin:
Updates the function list for MySQL using the latest version (8.0.27). Some of the added functions were already added as keywords (like `CUME_DIST`), but given that `CURRENT_DATE` is both a keyword and builtin function, figured they should be added here in similar fashion.

Documentation pages used:
- https://dev.mysql.com/doc/refman/8.0/en/loadable-function-reference.html
- https://dev.mysql.com/doc/refman/8.0/en/loadable-function-reference.html

For the second page, I only added a few of the functions on that page as I wasn't sure if all should be added including the superuser only ones. For reference, there were already a handful of functions that were added (like `ASYMMETRIC_*` ones), but others had not, so I just went with the new that I personally use from time to time, the `gen_*` and `mask_*` functions.


alexdima:
Thank you!

PR #2757, opened 2021-11-10
Update list of keywords for redshift

MasterOdin:
Updates the list of reserved keywords for Redshift based on https://docs.aws.amazon.com/redshift/latest/dg/r_pg_keywords.html.


hediet:
Thanks!

PR #2770, opened 2021-11-15
Expose the TypeScript object inside the ts worker

orta:
I've migrated our infra over to handle the new monorepo, but hit a blocker. In the older versions of monaco-typescript, importing `vs/language/typescript/tsWorker` would trigger `window.ts` to become the TypeScript object. 

This is _generally_ considered a  feature (of a side effect) when importing typescriptServices.js because of this `var`:

Screen Shot 2021-11-15 at 10 40 02 PM
 
I used that to access the same version of TypeScript as the monaco webworker in the Playground. 

I'm not tied to the implementation detail as it's easy to do both, so I felt this might be more ESM-y instead of jamming a `window.ts = ts` at the end but I think it depends on whether this is classed as a breaking change (and I guess if other people used that side-effect) in which case this can be switched out

### DIFF
--- a/src/typescript/tsWorker.ts
+++ b/src/typescript/tsWorker.ts
@@ -494,3 +494,6 @@ export function create(ctx: worker.IWorkerContext, createData: ICreateData): Typ
 
 	return new TSWorkerClass(ctx, createData);
 }
+
+/** Allows for clients to have access to the same version of TypeScript that the worker uses */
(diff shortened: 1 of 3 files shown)


orta:
I don't think this is enough to fix it

alexdima:
:+1: I've also added an assertion for it in the smoke test.

PR #2772, opened 2021-11-16
Remove return from example addAction

MasterOdin:
Documentation for [`editor.AddAction`](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.IActionDescriptor.html#run) gives a return type of `void | Promise<void>`. Having the example `return null` is confusing given it doesn't really matter what it returns.


alexdima:
Thank you!

PR #2806, opened 2021-12-08
Update Elixir tokenization of sigil modifiers

jonatanklosko:
Elixir has syntax for custom string interpreters like `~r/regex/i`. The trailing part is a set of modifiers, in this case making the regexp case-insensitive. The modifiers could only be lower and uppercase characters, but a recent change allows numbers too. See https://github.com/elixir-lang/elixir/issues/11282.

I also updated operators with the `**` (power), which has been added recently.

*Sidenote: when building the website locally it always used the old version of tokenizers, I even tried resetting the repo, running install, release, build-website, simpleserver and it would still pick up the old version. Is there an option to build from local files?*


hediet:
Thank you!

> Sidenote: when building the website locally it always used the old version of tokenizers, I even tried resetting the repo, running install, release, build-website, simpleserver and it would still pick up the old version. Is there an option to build from local files?

Please see here #2807

Issue #2700, opened 2021-10-22
Parse protobuf syntax incorrect without `syntax=` directive


gitpaladin:
In [protobuf language guide](https://developers.google.com/protocol-buffers/docs/proto3#simple), the `syntax` directive is optional. But a file without `syntax` is regarded as a invalid protobuf file in [src/protobuf/protobuf.ts](https://github.com/microsoft/monaco-languages/blob/7e1d706f5f42c54781fe0747c3d41871112d186a/src/protobuf/protobuf.ts#L126).

> The first line of the file specifies that you're using proto3 syntax: if you don't do this the protocol buffer compiler will assume you are using proto2. This must be the first non-empty, non-comment line of the file.

So the parser is not working without `syntax` declaration. (see below). 

![image](https://user-images.githubusercontent.com/8654635/138438776-24f4e14d-82f7-4428-abbe-35edd00afb4b.png)
  

**monaco-editor version:** 0.29.1
**Browser:** Edge 94.0.992.50 
**OS:** Win10 10.0.19043.1237
**Playground code that reproduces the issue:**

See in this [codepen](https://codepen.io/cmlchen/pen/dyzOmKO)

alexdima:
fixed via https://github.com/microsoft/monaco-editor/pull/2809

PR #2809, opened 2021-12-09
Fix token while missing `syntax=` directive

gitpaladin:
see this issue. https://github.com/microsoft/monaco-editor/issues/2700 Just simply `switchTo` proto2 token if there is no `sytanx=` provided.


hediet:
Thanks! Can you fix CI?

alexdima:
Thank you!

PR #2825, opened 2021-12-14
fix languages and features options for webpack plugin

yiminghe:
after build, do not include any other languages and features code

### DIFF
--- a/webpack-plugin/src/index.ts
+++ b/webpack-plugin/src/index.ts
@@ -312,7 +312,7 @@ function createLoaderRules(
 	};
 	return [
 		{
-			test: /esm[/\\]vs[/\\]editor[/\\]editor.(api|main).js/,
+			test: /esm[/\\]vs[/\\]editor[/\\]editor\.main\.js/,
 			use: [
 				{
 					loader: INCLUDE_LOADER_PATH,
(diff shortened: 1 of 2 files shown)


alexdima:
@yiminghe What is the purpose of this PR?

yiminghe:
reduce build size and speed:
```
new (require('monaco-editor-webpack-plugin'))({
          languages:['typescript']
        })
```

alexdima:
@yiminghe I cannot accept this PR. The problems are:
* the CI is broken by this proposed change
* the change is not motivated clearly, it is still not clear to me what the problem is.
* testing for `editor.api` is intentional, as we expect folks that want only a subset of the editor to import only that
* requiring `editor.all.js` will break shipping only selected features, as it will include all changes.

yiminghe:
This plugin is not working as expected when passing languages and features option.

I think you should try and compare built files before using this patch and after using this patch 

@alexdima

alexdima:
@yiminghe The plugin works correctly for me. For example, when I take the sample at https://github.com/microsoft/monaco-editor/tree/main/samples/browser-esm-webpack-monaco-plugin and modify the `webpack.config.js` to use something like:
```
	plugins: [
		new MonacoWebpackPlugin({
			languages: ['typescript', 'javascript', 'css'],
			features: ['hover']
		})
	]
``` 
I can observe in the generated files that the hover feature is present, while, for example, the find feature is not present.

yiminghe:
@alexdima checkout this: https://github.com/yiminghe/monaco-editor-2825

Issue #2823, opened 2021-12-13
Update Parcel install documentation...


resistdesign:
### Context

- [X] This issue is not a bug report. *(please use a different template for reporting a bug)*
- [X] This issue is not a duplicate of an existing issue. *(please use the [search](https://github.com/microsoft/monaco-editor/issues) to find existing issues)*

### Description

Here's a way better way to setup Monaco Editor for Parcel (2.0.1)...
DO NOT BUILD THE WORKERS IN A BASH SCRIPT...
Simply import the workers as URLs like this:

```tsx
// @ts-ignore
import JSONWorker from 'url:monaco-editor/esm/vs/language/json/json.worker.js';
// @ts-ignore
import CSSWorker from 'url:monaco-editor/esm/vs/language/css/css.worker.js';
// @ts-ignore
import HTMLWorker from 'url:monaco-editor/esm/vs/language/html/html.worker.js';
// @ts-ignore
import TSWorker from 'url:monaco-editor/esm/vs/language/typescript/ts.worker.js';
// @ts-ignore
import EditorWorker from 'url:monaco-editor/esm/vs/editor/editor.worker.js';
import React, { FC, HTMLAttributes, useEffect, useRef } from 'react';
import * as monaco from 'monaco-editor';
import styled from 'styled-components';

(window as any).MonacoEnvironment = {
  getWorkerUrl: function (_moduleId: string, label: string) {
    if (label === 'json') {
      return JSONWorker;
    }
    if (label === 'css' || label === 'scss' || label === 'less') {
      return CSSWorker;
    }
    if (label === 'html' || label === 'handlebars' || label === 'razor') {
      return HTMLWorker;
    }
    if (label === 'typescript' || label === 'javascript') {
      return TSWorker;
    }
    return EditorWorker;
  },
};

const Base = styled.div`
  min-width: 30em;
  min-height: 20em;
`;

export type MonacoEditorProp = HTMLAttributes<HTMLDivElement> & {
  value: string;
  language: string;
  onChange: (value: string) => void;
};

export const MonacoEditor: FC<MonacoEditorProp> = ({ value, language, onChange, ...other }) => {
  const baseRef = useRef<HTMLDivElement>(null);

  useEffect(() => {
    if (baseRef.current) {
      const newEditor = monaco.editor.create(baseRef.current as HTMLDivElement, {
        value,
        language,
      });
      const subscription = newEditor.onDidChangeModelContent(() => {
        onChange(newEditor.getValue());
[...]

PR #2829, opened 2021-12-15
Doc/simplify browser esm parcel

resistdesign:
A PR to update the Browser ESM Parcel sample as per: https://github.com/microsoft/monaco-editor/issues/2823#issuecomment-993706028


hediet:
Thank you for this PR!
It would be awesome if you could add `.idea/` to the ignore file in a separate PR though, so that the history reads nicely.

resistdesign:
Ok, I can do that later today 🙂

resistdesign:
Closing in favor of a cleaner PR.

alexdima:
Here's also a trick I use a lot in case I ever want to ignore a folder completely in a git repository without touching the top level `.gitignore`: just add a `.gitignore` file in the folder you want completely ignored with the content `*`.

resistdesign:
Whoa! Brilliant!

PR #2840, opened 2021-12-20
Add Cairo language

assimovt:
This PR adds support for [Cairo Language](https://www.cairo-lang.org/).

### DIFF
--- a/src/basic-languages/cairo/cairo.contribution.ts
+++ b/src/basic-languages/cairo/cairo.contribution.ts
@@ -1,0 +1,24 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { registerLanguage } from '../_.contribution';
+
+declare var AMD: any;
+declare var require: any;
+
+registerLanguage({
+	id: 'cairo',
+	extensions: ['.cairo'],
+	aliases: ['cairo', 'Cairo'],
+	loader: () => {
+		if (AMD) {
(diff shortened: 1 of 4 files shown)


hediet:
Thank you for your PR!

However, we try to limit the languages we bundle with the monaco editor and only accept languages of significant relevance (for example those that have a Wikipedia article).

The Cairo language seems to be quite new and I could not find a Wikipedia article about it, thus I suggest to wait adding built-in support for this language until it is more widely used.

Having said that, you can simply create an npm package that provides support for this language.
Have a look at [this demo](https://microsoft.github.io/monaco-editor/playground.html#extending-language-services-custom-languages) on how to register custom languages at runtime.

assimovt:
@hediet I see, thanks for your comment. Do you happen to have some links to a well-structured NPM package of some Monaco language extensions? It'd be great to provide some guidance for 3rd party language packages until they mature for the mainline.

hediet:
Unfortunately not, but I recommend the following structure:

```ts
export function registerCairoLanguageSupport(_monaco: typeof monaco) {
    _monaco.languages.setMonarchTokensProvider(...);
    _monaco.languages.registerCompletionItemProvider(...);
    ...
}
```

I.e. the npm package should expose a single function that registers all the providers/tokenizers etc. to the given monaco object.

assimovt:
@hediet Thanks for your help. I've created the repo here https://github.com/assimovt/monaco-language-cairo and also an NPM package. Any advice on how to port testing part w/o anything unnecessary from this repo? Ideally, I'd want to keep the same kind of .test file I created in PR and just run a minimized version of your test suite w/o requiring the monaco release.

hediet:
Unfortunately, we don't have the resources at the moment to provide advice for that.

If you manage to extract the testing functionality of this repo, I'm sure others would appreciate if you could document how to do that!

PR #2853, opened 2021-12-24
tune(monaco-editor-webpack-plugin): expose plugin options type

ValeraS:
Our build tools use this plugin under the hood, and its options are part of a larger configuration. So it would be great to get the types directly from this plugin.


hediet:
Thanks for this PR! I don't know the semantics of `declare namespace` by heart. What is the advantage of this over exporting the interface type?

ValeraS:
When using [`export = something`](https://github.com/microsoft/monaco-editor/blob/a385674a34a33947483e0f7266b5a7dbcefe9b9d/webpack-plugin/src/index.ts#L367), we cannot export anything else from this module.

alexdima:
Nice trick (to use a namespace with the same name as the class)!

PR #2865, opened 2021-12-30
tune(monaco-editor-webpack-plugin): replace editor.main.js

ValeraS:
Usually editor functionality used as:
```js
import * as monaco from 'monaco-editor';
``` 
this resolves to `./esm/vs/editor/editor.main.js` in [package.json](https://github.com/microsoft/monaco-editor/blob/25e85af033388fe4825418b76a89ef617199b069/package.json#L27). 

`editor.main.js` includes all feature and languages. Because of this, `webpack-plugin` does not work as expected.

This PR proposes replacing `editor.main.js` with `editor.api.js` and requested features and languages.


hediet:
Thanks for your PR! Can you please open an issue to describe the problem that this PR fixes?

ValeraS:
> Thanks for your PR! Can you please open an issue to describe the problem that this PR fixes?

Should I open an issue with text from the first comment or should I describe something more fully?

hediet:
Please also describe what you mean with "does not work as expected". Thank you!

ValeraS:
Closes #2874 
Maybe also need to change this line https://github.com/microsoft/monaco-editor/blob/a385674a34a33947483e0f7266b5a7dbcefe9b9d/webpack-plugin/src/index.ts#L315 to check only `editor.main.js`

hediet:
Great idea!

To test this change, we adopted the smoke-test to import `monaco-editor` (instead of `./esm/vs/editor/editor.api.js`), but unfortunately this breaks CI now.
Can you look into it?

Thank you!

webpack-plugin/src/index.ts
```diff
@@ -313,7 +313,7 @@ function createLoaderRules(
 	};
 	return [
 		{
-			test: /esm[/\\]vs[/\\]editor[/\\]editor.(api|main).js/,
+			test: /esm[/\\]vs[/\\]editor[/\\]editor.main.js/,
```
alexdima, line 316: I wonder, is this a breaking change for all existing users which were importing `'monaco-editor/esm/vs/editor/editor.api'`?

ValeraS: Yes, I commented it in commit message:

BREAKING CHANGES:
if a project contains only these imports for creating editor:
```
import * as monaco from 'monaco-editor/esm/vs/editor/editor.api.js';
const editor = monaco.editor.create(...);
```
`monaco-editor-webpack-plugin` will not add features and languages. When creating editor,
 monaco must be imported from package itself:
 ```
 import * as monaco from 'monaco-editor';
 const editor = monaco.editor.create(...);

hediet (approved):

yaohwu:
👍 👀

jmroon:
Any chance we can bring this in? We deploy with our own custom language and it's unfortunate to be forced to bundle 89 languages along with it.

stropalov:
We have the same issue, would be great to fix it.

vladplesu:
It would be great to have this merged so please don't forget about this PR.

aeschli (dismissed):

PR #2871, opened 2022-01-03
Fix #2851 Highlight correctly the attributes and identifiers (with dashes) for Shell language

### LINKED ISSUE #2851: [Bug] shell syntax colors the "-bar" part of "foo-bar" as an attribute

- [X] Not reproducible in [vscode.dev](https://vscode.dev) or VS Code Desktop

Use language=shell and this content:

```shell
foo-bar --baz gorp
```

### Actual Behavior

Screen Shot 2021-12-23 at 2 43 23 PM

### Expected Behavior

Screen Shot 2021-12-23 at 2 43 52 PM

### Additional Context

The relevant regexp is here: https://github.com/microsoft/monaco-editor/blob/main/src/basic-languages/shell/shell.ts#L162

Perhaps the regexp, instead of `/-+\w+/` should have a whitespace positive lookbehind? e.g. `(?<=\s)-+\w+`. I realize that this runs into problems with Safari and old IE... I am not familiar with the internals of monaco-editor, but perhaps there are already runtime guards in place that could be used here.

Or maybe something like `[/(\s)(-+[\w-]+/, ["white","attribute.name"]]`... but here my knowledge of monaco-editor source is too shallow at the moment to say with any certainty.

fwiw, vscode's markdown preview does not syntax color anything...
Screen Shot 2021-12-23 at 2 44 46 PM

juan-carlos-diaz:
This PR fix #2851. The fix is correcting the regex that highlight the attributes and create a new regex for identifiers containing dashes.

This impact the Monaco editor playground only.

You can test by running the tests:
>npm run release
>npm run test

Two new tests added, one for the bug reported and one for the new regex defined (identifiers with dashes).

### DIFF
--- a/src/basic-languages/shell/shell.test.ts
+++ b/src/basic-languages/shell/shell.test.ts
@@ -33,6 +33,40 @@ testTokenization('shell', [
 				{ startIndex: 14, type: 'white.shell' },
 				{ startIndex: 15, type: '' }
 			]
+		},
+		// Tests for case reported in bug #2851, do not confuse identifier(with dashes) with attribute
+		{
+			line: 'foo-bar --baz gorp -123 --abc-123',
+			tokens: [
+				{ startIndex: 0, type: '' },
+				{ startIndex: 7, type: 'white.shell' },
+				{ startIndex: 8, type: 'attribute.name.shell' },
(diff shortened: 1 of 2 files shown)


hediet:
Thanks for the PR!

PR #2921, opened 2022-01-26
Improved module specification in `package.json`

benmccann:
It's not really valid according to the spec to not have a `main` field. If you are distributing multiple versions like CJS and ESM you can specify one in `main` and the ESM in `module`, but you shouldn't ever really have a missing `main` field. If the main is an ESM file you can simply specify `"type": "module"`

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,8 @@
 		"watch": "tsc -w -p ./src"
 	},
 	"typings": "./esm/vs/editor/editor.api.d.ts",
-	"module": "./esm/vs/editor/editor.main.js",
+	"type": "module",
+	"main": "./esm/vs/editor/editor.main.js",
 	"repository": {
 		"type": "git",
 		"url": "https://github.com/microsoft/monaco-editor"


hediet:
Makes sense.

bluwy:
> It's not really valid according to the spec to not have a main field

I'm not sure if this is true. `main` should be CJS only, regardless of `type: module`. Nodejs only supports ESM entrypoints via the `exports` field. `module` is used by bundlers only.

benmccann:
> main should be CJS only, regardless of type: module

Really? Do you have a reference for that? The docs for `"type"` seem to align with how I believe it works https://nodejs.org/api/packages.html#type

I'll need to push an update to the PR though. Adding `"type": "module"` means we need to switch the build scripts to ESM

bluwy:
According to https://nodejs.org/api/packages.html#main, it's used for `require()` only, so having `main` as ESM doesn't sound right. I was actually working on a PR too to simply add a `package.json` of `type:module` into the release `/esm` folder 😬 maybe we can go with that

benmccann:
Those docs are rather misleading. I sent a PR to the Node docs to clarify: https://github.com/nodejs/node/pull/41720

But this PR would require too much work for me to proceed with. I'd need to convert the whole codebase to ESM to add `"type": "module"` and I didn't realize how many other places there's CJS code still, so I'll go ahead and close this

PR #2923, opened 2022-01-27
Ensure esm code is evaluated as esm

bluwy:
According to the [node spec](https://nodejs.org/api/packages.html#type), all `.js` file in a non `type:module` package.json is considered as CJS. This also includes the release `/esm`, which contains ESM code but with an extension of `.js`.

One way to fix this is to use:
1. `.mjs` for the files in `/esm`
2. Or have a `package.json` in the `/esm` folder to set all descendant `.js` as ESM (This PR)

I went with no2 as it's the least potential breaking change.

Note: This should give a clearer error message when one tries to import this package in node, even though it's browser-exclusive.


bluwy:
My bad. Looks like `/esm/metadata.js` is in CJS. I moved the package.json into `/esm/vs/` directory instead.

hediet:
What is your use-case of loading the monaco editor in nodejs?

bluwy:
There isn't really a usecase to load monaco editor in nodejs, but if it were to happen e.g. via a framework where client+server code can be in the same file (https://github.com/sveltejs/kit/discussions/3539), I think it should still silently work, or have an error that it doesn't work in nodejs.

Currently when loading in nodejs, we get the `Cannot use import statement outside a module` error instead because the ESM file from monaco editor is being treated as CJS.

I tried looking into the CI errors last week. Got a strange error with webpack still loading the ESM file with require. I saw that `main` have some great updates recently, I could perhaps revisit this again soon.

bluwy:
Dug more into this today, but I can't seem to figure out why the error is happening. With this PR, whenever I run `npm run package-for-smoketest --prefix webpack-plugin`, the final output in `test/smoke/webpack/out/app.js` for some reason contains `require('../something')` which wouldn't work in the browser. Without this PR, the `require` was transformed to `__webpack_require__` which works.

I'm guessing it has something to do with `monaco-editor-core` having the same extension issue as well, because the package is part of the release process, but I could be wrong. With that said, I don't think I have the bandwidth to proceed with this PR at the moment, so I'll close it. Feel free if anyone likes to tackle it though.

PR #2924, opened 2022-01-28
Add @preconcurrency to Swift declaration attributes

philipturner:
Citing [this pull request](https://github.com/apple/swift/pull/41022), the Core Swift team has decided on adding the new keyword to Swift 5.6. This will also ensure that syntax coloring for Swift-Colab is as up-to-date as possible if Google suddenly ups their monaco dependency from 0.27 to current.


hediet:
Thanks!

PR #2926, opened 2022-01-30
Only run publish workflow on main monaco-editor repo

MasterOdin:
Currently, the publish workflow will automatically run in any fork of the main repo, triggering an action failure once a week.

This PR adjusts the workflow so that it'll check if it's running in the main repo before attempting to run the job, so it'll continue to work the same in the main repo, and just not run at all in forks.

This could also be applied to the other workflows that run on cron jobs if desired.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -12,6 +12,7 @@ on:
         default: 'true'
 jobs:
   publish:
+    if: github.repository == "microsoft/monaco-editor"
     name: Publish to npm
     runs-on: ubuntu-latest
     steps:


hediet (approved):
Looks good, thanks!

hediet:
Hmm, the syntax seems to be different. I should have checked that.
https://github.com/microsoft/monaco-editor/actions/runs/1788088945/workflow

MasterOdin:
This was made when copying from the GH docs: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-run-job-for-specific-repository

I guess it only works if you use single quotes, or is the example in the docs wrong? 🤷

PR #2929, opened 2022-01-31
Add .cjs extension for javascript files

ZusorCode:
This extension is used by node projects of `"type": "module"` for files that are still written using CommonJS (for example configuration files).


hediet (approved):
Looks good, thank you!

Issue #2931, opened 2022-02-01
[Bug] typescriptDefaults is reinitialized after the creation of the first model


CGNonofr:
### Reproducible in vscode.dev or in VS Code Desktop?

- [X] Not reproducible in [vscode.dev](https://vscode.dev) or VS Code Desktop

### Reproducible in the monaco editor playground?

- [ ] Not reproducible in [the monaco editor playground](https://microsoft.github.io/monaco-editor/playground.html)

### Monaco Editor Playground Code

```typescript
const typescriptDefaults = monaco.languages.typescript.typescriptDefaults

monaco.editor.create(document.getElementById('container'), {
    model: monaco.editor.createModel(``, 'typescript')
});

setTimeout(() => {
    console.log(monaco.languages.typescript.typescriptDefaults === typescriptDefaults)
}, 1000)
```

### Actual Behavior

`false` is printed in the console

### Expected Behavior

`true` should be printed in the console

### Additional Context

This produces weird/unpredictable consequences related to the typescript worker when adding/removing extra libs

hediet:
I'm not so sure this qualifies as a bug, but it looks fishy indeed.

> This produces weird/unpredictable consequences related to the typescript worker when adding/removing extra libs

Can you elaborate on that?

CGNonofr:
Just to elaborate a little:

If you add an extra lib or set the typescript compiler options THEN create a model, the extra lib and the compiler options will be ignored

hediet:
I see, so not only does the reference change, but the new value is not even structurally equal to the old value?

CGNonofr:
I'm not sure what you mean by `structurally`

Here is a probably clearer bug highlight:
```typescript
const typescriptDefaults = monaco.languages.typescript.typescriptDefaults

typescriptDefaults.setCompilerOptions({
    esModuleInterop: true
})

const model = monaco.editor.createModel(
    `at the begining: ${typescriptDefaults.getCompilerOptions().esModuleInterop}\n`,
    'typescript',
    monaco.Uri.file('/test.ts')
)
monaco.editor.create(document.getElementById('container'), { model });

setTimeout(() => {
    model.setValue(model.getValue() + `after a model is created: ${typescriptDefaults.getCompilerOptions().esModuleInterop}\n`)
}, 1000)
```

hediet:
Fixed by 69809452a843de346958f253de730d567994debb.

PR #2932, opened 2022-02-01
Properly mark files as external

CGNonofr:
Without it, some code is duplicated at the end.

The creation of the `LanguageServiceDefaultsImpl` are in both `monaco.contribution.js` AND `tsMode.js`

The first time the `typescript` language is used, `tsMode` is dynamically imported from `monaco.contribution`, which reset the `typescriptDefaults`/`javascriptDefaults` by creating new instances of `LanguageServiceDefaultsImpl`

fix for https://github.com/microsoft/monaco-editor/issues/2931


hediet:
Fixed in https://github.com/microsoft/monaco-editor/commit/69809452a843de346958f253de730d567994debb.

PR #2957, opened 2022-02-10
Add `@noDerivative` modifier to Swift

philipturner:
This is part of the AutoDiff modifiers, alongside `@derivative` and `@differentiable`. I forgot to add it in a previous commit. It's used extensively in differentiable Swift, for example [here](https://github.com/tensorflow/swift-apis/blob/f51ee4618d652a2419e998bf9418ad80bda67454/Sources/TensorFlow/Layers/Dropout.swift#L40).


hediet:
Thanks!

PR #2968, opened 2022-02-14
Add browser-script-vite-vue sample

junyiz:


samples/browser-script-vite-vue/index.html
```diff
@@ -0,0 +1,21 @@
...
+		<meta name="viewport" content="width=device-width, initial-scale=1.0" />
+		<title>Vite App</title>
+		<link
+			rel="stylesheet"
+			data-name="vs/editor/editor.main"
+			href="https://unpkg.com/monaco-editor@0.32.1/min/vs/editor/editor.main.css"
+		/>
+		<script src="https://unpkg.com/monaco-editor@0.32.1/min/vs/loader.js"></script>
+		<script src="https://unpkg.com/monaco-editor@0.32.1/min/vs/editor/editor.main.nls.js"></script>
+		<script src="https://unpkg.com/monaco-editor@0.32.1/min/vs/editor/editor.main.js"></script>
```
hediet, line 15: Hmm, I'm not sure if this is a good way to load the monaco editor. Ideally the monaco editor is distributed locally and not loaded from a CDN. Examples should not encourage to do that.

hediet:
Thanks for this PR!
Unfortunately, nobody in our team has experience with Vue and Vite, so it will be hard for us to maintain this example.

Also, I'm not sure about loading the editor from a CDN, as the example dies when the CDN goes down.

PR #3001, opened 2022-03-01
Upgrade ts-node

andrewbranch:
monaco-typescript builds of TypeScript 4.6 and nightlies are failing due to a TS API change that ts-node needed to account for: https://github.com/microsoft/TypeScript/issues/47921 / https://github.com/TypeStrong/ts-node/pull/1648. The ts-node fix was just published to v10.6. We need this in order to get new versions of TypeScript up on the playground. Thanks!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,7 +27,7 @@
 				"pretty-quick": "^3.1.3",
 				"requirejs": "^2.3.6",
 				"terser": "^5.10.0",
-				"ts-node": "^10.4.0",
+				"ts-node": "^10.6.0",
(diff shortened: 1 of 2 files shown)


hediet:
Thanks for the PR!

However, I find it strange that a build-time dependency to `ts-node` can cause such problems.
Is this because `ts-node` requires `typescript` as peer-dependency and there was a breaking change in a typescript version that is compatible with `ts-node`'s version constraint of that peer-dependency?

andrewbranch:
@hediet exactly. ts-node has a peer dep of typescript@>=2.7, and we had an API breaking change.

PR #3023, opened 2022-03-15
Add new required methods to LanguageServiceHost

andrewbranch:
The comment from @weswigham on these methods says

```
/*
 * Unlike `realpath and `readDirectory`, `readFile` and `fileExists` are now _required_
 * to properly acquire and setup source files under module: node12+ modes.
 */
```

Obviously monaco-editor can’t read files, so I assume this is the closest we’re going to get. I’m not sure when/if these methods will get called, but the Playground does let you select node12/nodenext module settings, so I think throwing a not-implemented error would likely show up. @weswigham can you review please?


weswigham (approved):
Yeah, this is about what I'd expect the implementations to be for a more VFS based host.

hediet:
Thanks!

Issue #2295, opened 2021-01-02
Models with "@" in their name do not resolve as dependencies


lukasbach:
The following code examples can be copied into https://microsoft.github.io/monaco-editor/playground.html for verification.

I'm trying to specify dependencies manually via ``monaco.editor.createModel()``, which works fine for dependency outside of organizations, but if a package is defined within a organization, i.e. has a name like ``@org/packagename`` and thus contains a @-character, it breaks. To me it looks like this is due to Monaco changing the @-character to ``%40``.

The following code (which does not contain @) works fine:

```js
monaco.editor.create(document.getElementById("container"), {
    model: monaco.editor.createModel(
        "import * as t from 'test';\nt.X.toExponential();\nt.Y; // Error", 
        "typescript",
        monaco.Uri.parse("file://root/test.ts")
    )
});

monaco.languages.typescript.typescriptDefaults.setCompilerOptions({
  ...monaco.languages.typescript.typescriptDefaults.getCompilerOptions(),
  moduleResolution: monaco.languages.typescript.ModuleResolutionKind.NodeJs
});

monaco.editor.createModel(
    JSON.stringify({
        name: "test", version: "1.0.0",
        typings: "./lib/index.d.ts"
    }), 
    "typescript",
    monaco.Uri.parse("file://root/node_modules/test/package.json")
);

monaco.editor.createModel(
    "export const X = 1;", 
    "typescript",
    monaco.Uri.parse("file://root/node_modules/test/lib/index.d.ts")
);

console.log(
    "Works:",
    monaco.editor.getModel("file://root/node_modules/test/lib/index.d.ts").uri.toString()
);
```

However, on this code (containing a @) it breaks;

```js
monaco.editor.create(document.getElementById("container"), {
    model: monaco.editor.createModel(
        "import * as t from '@org/test';\nt.X.toExponential();\nt.Y; // Error", 
        "typescript",
        monaco.Uri.parse("file://root/test.ts")
    )
});

monaco.languages.typescript.typescriptDefaults.setCompilerOptions({
  ...monaco.languages.typescript.typescriptDefaults.getCompilerOptions(),
  moduleResolution: monaco.languages.typescript.ModuleResolutionKind.NodeJs
});

monaco.editor.createModel(
    JSON.stringify({
        name: "@org/test", version: "1.0.0",
        typings: "./lib/index.d.ts"
    }), 
    "typescript",
    monaco.Uri.parse("file://root/node_modules/@org/test/package.json")
[...]

PR #3057, opened 2022-04-04
Fix issue #2295 - Models with "@" in their name do not resolve as dependencies

bsorrentino:
fix issue #2295

component: `TypeScriptWorker` (`tsWorker.ts`)

- Enable '**skipEncoding**' flag on `Uri.toString` invokation on `getScriptFileNames()` method
- Compare '**fileName**' argument provided to `_getModel()` method both with Uri encoded and not

### DIFF
--- a/src/language/typescript/tsWorker.ts
+++ b/src/language/typescript/tsWorker.ts
@@ -63,14 +63,15 @@ export class TypeScriptWorker implements ts.LanguageServiceHost, ITypeScriptWork
 
 	getScriptFileNames(): string[] {
 		const allModels = this._ctx.getMirrorModels().map((model) => model.uri);
-		const models = allModels.filter((uri) => !fileNameIsLib(uri)).map((uri) => uri.toString());
+		const models = allModels.filter((uri) => !fileNameIsLib(uri)).map((uri) => uri.toString(true));
 		return models.concat(Object.keys(this._extraLibs));
 	}
 
 	private _getModel(fileName: string): worker.IMirrorModel | null {
 		let models = this._ctx.getMirrorModels();
 		for (let i = 0; i < models.length; i++) {
-			if (models[i].uri.toString() === fileName) {
+			const uri = models[i].uri;
+			if (uri.toString() === fileName || uri.toString(true) === fileName) {
 				return models[i];
 			}
 		}


hediet:
Thanks for the PR and your investigation!

I don't fully understand the problem though, why does this fix the issue?

bsorrentino:
> Thanks for the PR and your investigation!
> 
> I don't fully understand the problem though, why does this fix the issue?

Hi @hediet thanks for starting review

I tackled issue #2295 integrating monaco editor in my app.

As you could see from issue's comments the problem was related to typescript language when we import the scoped package (@scope/package)

@lstkz has figure out that using %40 instead of @ the problem disappear ( e.g. using %40azure/msal instead of @azure/msal` ) as states in this comment

With this hint I've investigated on the part that resolve the import and finally I got a fix.

Hope this clarify bit more

hediet:
That I understand, by I don't understand the underlying issue.

I think the actual issue is that it is unclear how `fileName` should be encoded in `private _getModel(fileName: string)`.

In this PR, I don't like the line `if (uri.toString() === fileName || uri.toString(true) === fileName) {` very much. Why do we need to check for both?

bsorrentino:
> That I understand, by I don't understand the underlying issue.
> 
> I think the actual issue is that it is unclear how `fileName` should be encoded in `private _getModel(fileName: string)`.
> 
> In this PR, I don't like the line `if (uri.toString() === fileName || uri.toString(true) === fileName) {` very much. Why do we need to check for both?

Hi @hediet 

tthe reason is that `uri.toString()` by default encode uri content and , for **scoped packages**, the `@` char is converted to `%40` and, in such case, the condition never match.

I've added the condition in `OR` to enforce backward compatibility

pietrovismara:
Would be awesome to get this merged...

peterp:
@bsorrentino Throwing my two cents in here because this seems to be blocked or ignored!

I don't think the code clearly communicates _why_ the two conditional checks are important, and your comment clarified the reasoning, but I get the sense that it would help to add that as a comment so that future readers of the code also understand why this was required.

Perhaps that may give @hediet a bit more confidence to merge this.

bsorrentino:
> @bsorrentino Throwing my two cents in here because this seems to be blocked.
> 
> I don't think the code clearly communicates _why_ the two conditional checks are important, and your comment clarified the reasoning, but I get the sense that it would help to add that as a comment so that future readers of the code also understand why this was required.
> 
> Perhaps that may give @hediet or @alexdima a bit more confidence to merge this.

Hi @peterp the suggestion here is to renew the PR adding meaningful comments to the affected code ?

hrgdavor:
cmon. why is this PR still blocked ?

hediet (approved):
After some exploration, I also don't have a better idea than the change suggested in this PR.

bpasero (approved):

PR #3058, opened 2022-04-05
Add yaml indent action

remcohaszing:
If a line ends with `:` and a user presses Enter, they are probably defining an object, which requires indentation.

[This rule is taken from `monaco-yaml`](https://github.com/remcohaszing/monaco-yaml/blob/main/src/yamlMode.ts#L41-L45), but I intend to remove basic language support there, since that’s already part of `monaco-editor`.


hediet:
I think this is reasonable. Thank you for the PR!

PR #3116, opened 2022-05-20
feat: add clickhouse language

MasterOdin:
Adds syntax support for the [ClickHouse](https://clickhouse.com/) database engine ([wikipedia page](https://en.wikipedia.org/wiki/ClickHouse)). The keywords are similar to other engines and follow the SQL standard, while the functions are all unique to clickhouse.


hediet:
Thanks for the PR! Same as for #3139:
Can you give us some context about this language and its relevance? I never heard of it before. How many people use it?
I wonder if it makes sense to ship this language with the monaco editor or if it should be distributed in its own npm package.

MasterOdin:
Closing this as probably won't be accepted, and better to make it its own npm package per the discussion in https://github.com/microsoft/monaco-editor/pull/3139

PR #3127, opened 2022-06-02
Fix false-positive check in `provideCompletionItems` method

Strate:
If entry.kindModifiers is undefined, then whole condition was true (because undefined !== -1), causing marking entry as deprecated.

### DIFF
--- a/src/language/typescript/languageFeatures.ts
+++ b/src/language/typescript/languageFeatures.ts
@@ -474,7 +474,7 @@ export class SuggestAdapter extends Adapter implements languages.CompletionItemP
 			}
 
 			const tags: languages.CompletionItemTag[] = [];
-			if (entry.kindModifiers?.indexOf('deprecated') !== -1) {
+			if (entry.kindModifiers && entry.kindModifiers.indexOf('deprecated') >= 0) {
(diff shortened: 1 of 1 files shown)


src/language/typescript/languageFeatures.ts
```diff
@@ -474,7 +474,7 @@ export class SuggestAdapter extends Adapter implements languages.CompletionItemP
 			}
 
 			const tags: languages.CompletionItemTag[] = [];
-			if (entry.kindModifiers?.indexOf('deprecated') !== -1) {
+			if (entry.kindModifiers && entry.kindModifiers.indexOf('deprecated') >= 0) {
```
hediet, line 477:
```suggestion
			if (entry.kindModifiers && entry.kindModifiers.indexOf('deprecated') !== -1) {
```

`indexOf(...) !== -1` is a common pattern.

hediet (changes requested):
See comments

hediet (approved):

hediet:
Thanks for the bug report!
Closing this in favor of https://github.com/microsoft/monaco-editor/pull/3531.

PR #3139, opened 2022-06-09
feat: add bigquery language

MasterOdin:
Adds syntax support for the [BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/introduction) database engine ([wikipedia page](https://en.wikipedia.org/wiki/BigQuery)).


hediet:
Can you give us some context about this language? I never heard of it before. How many people use it?
I wonder if it makes sense to ship this language with the monaco editor or if it should be distributed in its own npm package.

MasterOdin:
> Can you give us some context about this language and its relevance? I never heard of it before. How many people use it?

This is the language for BigQuery, which is Google's data warehouse (similar to Redshift from Amazon, which has a language file here). According to https://db-engines.com/en/ranking, it is one of the more commonly used data warehouses (below snowflake and databricks, but above redshift). According to https://cloud.google.com/customers#/products=Data_Analytics, its customers include Home Depot, Twitter, Paypal, UPS, etc.

> I wonder if it makes sense to ship this language with the monaco editor or if it should be distributed in its own npm package.

I'm fine with making its own package, though I'd ask that a wiki page (linked from the README?) be created to list these community languages to help with discoverability of these separate packages.

MasterOdin:
> I'd prefer an external package here, as the language itself seems to be tied to the product (or rather: the wikipedia page about the language is actually about the product).

With the exception of plain SQL and T-SQL, this statement is true of most all other SQL dialects as they're tied to specific DB engine. Like, there's no wikipedia page for the language postgres, redshift, or mysql uses. Same with redis.

I'm fine with the decision here to have it be separate, I'm just seeking clarification if there'd be a case for any additional SQL dialect be added as a basic-langauge.

MasterOdin:
> > though I'd ask that a wiki page (linked from the README?) be created to list these community languages to help with discoverability of these separate packages.
>
> Sure! Please create a PR for that!
> We can also list them in the readme.

It looks like only maintainers can update the wiki, so I think just listing them in the README makes sense then.

PR #3242, opened 2022-08-11
Switch to `yarn` which supports overriding resolutions to address vulnerabilities in dependencies of dependencies

alexdima:


remcohaszing:
This is supported by npm too, but they call it [`overrides`](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides). Not that I mind Monaco using yarn or npm.

lramos15 (approved):

silverwind:
Good decision. Yarn v1 shouldn't really be used anymore as it's long unmaintained, and the newer versions are just odd in many regards. 😉

PR #3243, opened 2022-08-13
Use monaco-editor-core final 0.34 release

CGNonofr:
monaco-editor 0.34 is not using monaco-editor-core 0.34

Also, the package-lock doesn't contain the right package version

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "monaco-editor",
-	"version": "0.33.0",
+	"version": "0.34.0",
 	"lockfileVersion": 2,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "monaco-editor",
-			"version": "0.33.0",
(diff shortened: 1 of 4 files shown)


CGNonofr:
Is there anything blocking this change from being merged? @hediet maybe?

alex-ander-is (approved):

alex-ander-is:
@alexdima We are missing regular versioning and correct points, which makes our dev life harder. Could you take a look please? Ideally incl. corresponding tag, please? Thank you!

hediet:
This is as intended. Both the stable and the nightly build will override the monaco-editor-core version in package.json.
The monaco-editor-core version in package.json is only relevant when checking out and working with the monaco-editor repository.

For the stable build, we build a fresh core package that is extracted from the vscode ref that is specified in package.json.
For nightly, we use the VS Code main branch.

PR #3283, opened 2022-09-01
Cleanup/reduce build complexity

dyst5422:


dyst5422:
Entirely just experimenting to figure out how to build a worker. Didn't mean to PR here.

PR #3352, opened 2022-10-05
Use transformer to remove require calls and nodeSystem from typescript.js

jakebailey:
This includes #3344; I would merge that in first and then I'll rebase this one as it's more complicated.

Right now, the build script uses source level hacks to remove require calls and such. But with https://github.com/microsoft/TypeScript/issues/49332 on the horizon, these source level hacks are going to be unstable.

Rather than try and maintain all of these regex replacements, rewrite the code using transpileModule. This should be a sure-fire method as we can use the helper TS uses to identify require calls and replace them with no-ops.

I've removed the part of the script which complains about leftover requires, as the transformer gets all of the "real" requires; anything left is a false positive from a string, or a call to `sys.require` which the transformer will also ensure is undefined as it removes the nodeSystem from use.

### DIFF
--- a/build/importTypescript.ts
+++ b/build/importTypescript.ts
@@ -35,9 +35,7 @@ const TYPESCRIPT_LIB_DESTINATION = path.join(REPO_ROOT, 'src/language/typescript
 export const typescriptVersion = "${typeScriptDependencyVersion}";\n`
 	);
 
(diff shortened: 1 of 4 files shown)


jakebailey:
Closing; #3356 is better.

PR #3356, opened 2022-10-11
Remove regex replacements from import-typescript

jakebailey:
Rather than modifying TS's source to try and remove require/module/debugger/etc, we can instead define those variables at the top and leave the source unmodified (which TS can handle), and then remove `debugger` at `esbuild` time.

This approach is a lot more flexible for TypeScript. We don't have to think about which random bits of code happen to be regex replaced here, and it could even mean that our package could be shipped minified or optimized in some way, if we end up doing that. Mainly, this means we don't have to add a whole slew of new special cases for the module-ified TS; this approach works for all modern versions of TS.

Supercedes #3352.

This includes #3344.


build/importTypescript.ts
```diff
@@ -118,15 +63,6 @@ define("vs/language/typescript/lib/typescriptServices", [], function() { return
...
 	);
 
-	// Remove pattern that creates warnings with esbuild
-	// e.g.
-	// > /src/typescript/lib/typescriptServices.js:20:21: warning: Top-level "this" will be replaced with undefined since this file is an ECMAScript module
-	// 20 │ var __spreadArray = (this && this.__spreadArray) || function (to, from, pack) {
-	//    ╵                      ~~~~
-	//
-
-	tsServices = tsServices.replace(/\nvar ([^ ]+) = \(this && this\.([^)]+)\) \|\|/gm, '\nvar $1 =');
```
jakebailey, line 128: This does have a runtime effect, but as far as I can tell esbuild doesn't complain when I run `npm run release`.

jakebailey:
This PR is now running on TypeScript-Make-Monaco-Builds and is working.

jakebailey:
> The reason we removed require calls from the file was because webpack and other bundlers were picking them up and inlining nodejs shims in the resulting bundle.

FWIW this shouldn't actually end up mattering for TypeScript internally; we detect that it's not a node system and never actually try and use those shims, if they weren't present. Of course, it's not so great that the bundle size would increase and/or warnings would appear.

alexdima (approved):

jakebailey:
Thanks!

PR #3334, opened 2022-09-27
Update @types/node

andrewbranch:
There was a compatibility issue with TypeScript 4.9 Beta in @types/node that was fixed in a recent version. This unblocks TypeScript Playground builds of the beta. Thanks!

PR #3389, opened 2022-10-28
Update node types for TS 4.9 RC

sandersn:
TS 4.9 RC updates AbortSignal again, and requires a matching upgrade to `@types/node`. Followup to #3334. Required to compile against TS 4.9 RC or Typescript nightly builds.


jakebailey:
cc @alexdima (this is breaking playground builds, unfortunately)

PR #3474, opened 2022-12-24
Add Vue3 sample.

recallwei:
- Add vite-vue3-javascript sample.
- Add vite-vue3-typescript sample.

### DIFF
--- a/samples/browser-esm-vite-vue3/src/components/Editor.vue
+++ b/samples/browser-esm-vite-vue3/src/components/Editor.vue
@@ -1,0 +1,29 @@
+<script setup>
+import { ref, onMounted, onUnmounted, toRaw } from 'vue';
+import * as monaco from 'monaco-editor/esm/vs/editor/editor.api';
+
+const editorRef = ref(null);
+const editorInstance = ref(null);
+
+onMounted(() => {
+	if (editorRef.value && !editorInstance.value) {
+		editorInstance.value = monaco.editor.create(editorRef.value, {
+			value: ['function x() {', '\tconsole.log("Hello world!");', '}'].join('\n'),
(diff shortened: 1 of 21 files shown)


samples/browser-esm-vite-vue3/src/components/Editor.vue
```diff
@@ -0,0 +1,29 @@
...
+
+onMounted(() => {
+	if (editorRef.value && !editorInstance.value) {
+		editorInstance.value = monaco.editor.create(editorRef.value, {
+			value: ['function x() {', '\tconsole.log("Hello world!");', '}'].join('\n'),
+			language: 'javascript'
+		});
+	}
+});
+
```
re2005, line 16: Would be nice to add a `save` example, witch in my case took me some work to find out how I should actually do it.

```ts
    const save = () => {
        const code = monaco.editor.getEditors()[0].getValue();
    };
```

recallwei: If you want to do this you must make sure the editor is instantiated, otherwise `getEditor()` may return `undefined`, which will cause an error.

re2005: @recallwei thanks. My suggestion is to add to this PR the example, so will be easier to other devs.

Cheers

recallwei:
@hediet

hediet:
Since our team is not familiar with Vue, we cannot maintain this sample in our repository.
Instead, you can link your sample from the readme.md in the sample directory.
Thanks for understanding!

re2005:
A bit of a shame here, this is open source and it should be open for people like us to contribute ?

PR #3508, opened 2023-01-19
Add inlay hints provider example

ununian:


ununian:

ununian:

hediet:
I don't think adding this example brings much value, also it seems like this example is already there. Please create an issue first!

PR #3533, opened 2023-02-07
Update info needed closer

lramos15:
With the label change this no longer worked, updates it to use the latest label


hediet (approved):

hediet:
Thanks!

PR #3554, opened 2023-02-13
Update @types/node

andrewbranch:
Playground builds are broken again with an `AbortSignal` type error, and I’m hazarding a guess this fixes it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1809,9 +1809,9 @@
 			"dev": true
 		},
 		"node_modules/@types/node": {
-			"version": "17.0.14",
-			"resolved": "https://registry.npmjs.org/@types/node/-/node-17.0.14.tgz",
-			"integrity": "sha512-SbjLmERksKOGzWzPNuW7fJM7fk3YXVTFiZWB/Hs99gwhk+/dnrQRPBQjPW9aO+fi1tAffi9PrwFvsmOKmDTyng=="
+			"version": "18.13.0",
+			"resolved": "https://registry.npmjs.org/@types/node/-/node-18.13.0.tgz",
+			"integrity": "sha512-gC3TazRzGoOnoKAhUx+Q0t8S9Tzs74z7m0ipwGpSqQrleP14hKxP4/JUeEQcD3W1/aIpnWl8pHowI7WokuZpXg=="
 		},
 		"node_modules/@types/parse-json": {
 			"version": "4.0.0",
@@ -8437,9 +8437,9 @@
 			"dev": true
 		},
 		"@types/node": {
-			"version": "17.0.14",
-			"resolved": "https://registry.npmjs.org/@types/node/-/node-17.0.14.tgz",
-			"integrity": "sha512-SbjLmERksKOGzWzPNuW7fJM7fk3YXVTFiZWB/Hs99gwhk+/dnrQRPBQjPW9aO+fi1tAffi9PrwFvsmOKmDTyng=="
+			"version": "18.13.0",
+			"resolved": "https://registry.npmjs.org/@types/node/-/node-18.13.0.tgz",
+			"integrity": "sha512-gC3TazRzGoOnoKAhUx+Q0t8S9Tzs74z7m0ipwGpSqQrleP14hKxP4/JUeEQcD3W1/aIpnWl8pHowI7WokuZpXg=="
 		},
 		"@types/parse-json": {
 			"version": "4.0.0",


hediet:
Thanks for the PR! Why does this break your playground build though?

andrewbranch:
The playground is built by building monaco from source in a git submodule combined with the version of TypeScript it hosts and other playground-specific code. From time to time, changes in TypeScript itself mean either monaco's API usage of TypeScript or global types that are present in the combined compilation have to be updated. Because we build against nightly versions of TypeScript, it usually means these errors are the same ones you would get in a few months when you pick up a new stable version of TypeScript—but instead, we have to fix them right away to get the playground working. You can see my [other PRs](https://github.com/microsoft/monaco-editor/pulls?q=is%3Apr+author%3Aandrewbranch+is%3Aclosed) for examples of this.

The error we’re currently getting in the playground builds is a conflict of the global type `AbortSignal`, which is defined both in `lib.dom.d.ts` and `@types/node`. These definitions have to agree to coexist in a compilation. My last PR to monaco, https://github.com/microsoft/monaco-editor/pull/3334, had the exact same symptom. This time, though, I did a quick look for what exactly caused the break, and I’m honestly not sure after a few minutes of investigating—I don’t see any recent changes to `AbortSignal`. However, keeping `@types/node` updated in your build should be harmless, so I decided that would be a good first step before investigating more. If you have concerns, though, I can do a more thorough investigation, or pull in a patch to the playground build.

andrewbranch:
The playground is built by building monaco from source in a git submodule combined with the version of TypeScript it hosts and other playground-specific code. From time to time, changes in TypeScript itself mean either monaco's API usage of TypeScript or global types that are present in the combined compilation have to be updated. Because we build against nightly versions of TypeScript, it usually means these errors are the same ones you would get in a few months when you pick up a new stable version of TypeScript—but instead, we have to fix them right away to get the playground working. You can see my [other PRs](https://github.com/microsoft/monaco-editor/pulls?q=is%3Apr+author%3Aandrewbranch+is%3Aclosed) for examples of this.

The error we’re currently getting in the playground builds is a conflict of the global type `AbortSignal`, which is defined both in `lib.dom.d.ts` and `@types/node`. These definitions have to agree to coexist in a compilation. My last PR to monaco, https://github.com/microsoft/monaco-editor/pull/3334, had the exact same symptom. This time, though, I did a quick look for what exactly caused the break, and I’m honestly not sure after a few minutes of investigating—I don’t see any recent changes to `AbortSignal`. However, keeping `@types/node` updated in your build should be harmless, so I decided that would be a good first step before investigating more. If you have concerns, though, I can do a more thorough investigation, or pull in a patch to the playground build.

hediet:
> The playground is built by building monaco from source in a git submodule combined with the version of TypeScript it hosts and other playground-specific code

Hmm, I'm not so happy that you have to build monaco from source for the typescript playground.

What can we do so that you can just consume the monaco-editor/monaco-editor-core npm package?
Ideally in a way that users of the monaco editor can also get an up-to-date TypeScript experience and that users can switch between TypeScript versions?

andrewbranch:
We’re very interested in doing this, and I think we plan to talk about it next week at our weekly meeting with @mjbvz. @jakebailey currently knows the most about this topic on our end.

In the meantime, we’re unblocked because Jake pulled this PR in as a patch in playground builds—but that does mean your build will fail sometime in the future when you update TypeScript in this repo, and this PR will fix it.

jakebailey:
Ideally, we'd be able to pass in the TS API as an object at runtime, though that does have implications for other downstream users of monaco in that right now, they get TS all bundled for free.

(Even better would be to use tsserver for this, but my understanding is that it isn't really doable given how much code is in the vscode typescript extension.)

But yes, we should meet about this.

andrewbranch:
> how much code is in the vscode typescript extension

If monaco-editor is an LSP client, we might be able to use the LSP/TSServer bridge built into the TS support for VS. I believe it’s less complete than the VS Code / TSServer integration, but maybe complete enough for the playground, and maybe portable?

hediet:
> If monaco-editor is an LSP client, we might be able to use the LSP/TSServer bridge built into the TS support for VS.

That would be cool.

> though that does have implications for other downstream users of monaco in that right now, they get TS all bundled for free.

They could use monaco-editor-core, monaco-editor could still ship some fixed typescript version.

chrmarti (approved):

hediet (approved):

PR #3568, opened 2023-02-16
Recreate package-lock.json to unblock TypeScript playground builds

jakebailey:
Builds of the TypeScript playground are failing because this package is referring to `@types/node@17`, a package that is not maintained as it was for an odd version. See: https://github.com/microsoft/TypeScript-Make-Monaco-Builds/actions/runs/4191887031/jobs/7266794864

This PR relocks `package-lock.json`, bumping everything to the latest compatible so we get the latest types too, which is needed to build with TS 5.0 and so on.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -157,25 +157,26 @@
 				"node": ">=4"
 			}
 		},
-		"node_modules/@cspotcode/source-map-consumer": {
-			"version": "0.8.0",
-			"resolved": "https://registry.npmjs.org/@cspotcode/source-map-consumer/-/source-map-consumer-0.8.0.tgz",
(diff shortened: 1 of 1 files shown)


jakebailey:
I went for the bulk hammer approach of just relocking it, but if this is not palatable, I can just try and fix the subset of deps blocking TS 5.0.

jakebailey:
For now I've done this on the playground repo, though: https://github.com/microsoft/TypeScript-Make-Monaco-Builds/commit/a29e648645a74c106f0f2ce5836ab17f14111dbf

So, that might be sufficient too.

jakebailey:
#3554, actually.

PR #3573, opened 2023-02-17
Update README Playground Link

ObieMunoz:
The link in the README to try the editor is dead.

Updated with a proper link to the playground.


hediet:
Thanks for reporting! Your link includes version, so I fixed it in https://github.com/microsoft/monaco-editor/pull/3575.

PR #3618, opened 2023-03-01
Update CONTRIBUTING.md

PmcFizz:
modify Running the website locally document

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -103,17 +103,16 @@ To debug bundled languages, such as JSON, HTML or TypeScript/JavaScript.
 ## Running the website locally
 
 ```bash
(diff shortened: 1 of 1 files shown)


PmcFizz:
@hediet  would you like check this mr？

hediet (approved):

hediet:
Please fix CI. Also, please remove the URL, as webpack might chose a different port. And it should be "Now webpack logs ..."

Thanks!

hediet (changes requested):
The top level package manager is npm, not yarn. So the first two commands should use npm.
Also, please use prettier to format

aeschli (approved):

hediet (approved):

PR #3619, opened 2023-03-01
Replace occurrences with highlights

jakebailey:
We just removed this in https://github.com/microsoft/TypeScript/pull/52347 but now we can't build the playground (oops).

Somehow we didn't find this use. We have had document highlights for a good while, so we can use it here.


mjbvz (approved):

jakebailey:
FWIW I checked and for the oldest version of TS that the TS playground supports (3.3), occurrences was already deprecated in favor of document highlights, so I believe this should be fairly safe.

hediet (approved):

PR #3720, opened 2023-03-03
Add a `customTypeScriptPath` option which should point to a script which defines a global `ts`

alexdima:
Allows configuring:
```ts
monaco.languages.typescript.typescriptDefaults.setWorkerOptions({
	customTypeScriptPath: 'https://someplace/typescript.js' // a file which will be loaded in the worker and which should define a global `ts` symbol
});
```


hediet:
Closing as there does not seem to be interest in this approach. Let's reopen once there is more interest in this.

PR #3350, opened 2022-10-04
Update typescript to 4.8.4

SferaDev:
Ideally `monaco` should allow customising the TypeScript version, 4.5.5 has several months and newer features are not available.

In the jump to 4.8, `InlayHintsOptions` has been removed and merged inside `UserPreferences`, not sure if you want to build a sub-type.

PR #3721, opened 2023-03-03
Update TypeScript to TS 5.0.2 in all projects, vendored files

jakebailey:
I'm working on making TS dynamically included. I noticed that `npm run import-typescript` wasn't run after TS was bumped to TS 4.7 which is making that change a little noisy because my change cleans up a bit of the vendored code (which I then can't commit because what's in main is the wrong files already).

I could send a PR that just finishes updating the vendored copy to what's defined in `package.json`, but it seems easiest to also just jump to the latest version of TypeScript at the same time.

This effectively supercedes #3350.


jakebailey:
Now updated to TS 5.0.

hediet:
Thanks!

aeschli (approved):

PR #3726, opened 2023-03-06
opt example 

PmcFizz:
when use iframe and the 'allow-modals' keyword is not set, the `alert` function is not used .


hediet (approved):

hediet:
Maybe it makes sense to set `allow-modals` though

ulugbekna (approved):

PR #3850, opened 2023-03-19
feat: ATA support

tamayika:
I added ATA(Automatic Type Acquisition) feature using `@typescript/ata`(https://github.com/microsoft/TypeScript-Website/tree/v2/packages/ata).

* default is disabled, so there is no downgrades.
* fetch type definition from jsdelivr.
* fetch is called in web worker.
  * this is because monaco.contribution.ts does not contains ts namespace and is keeping small size.
  * but web worker does not know when code is updated, so ATA runs in getScriptText method asynchronously.

I don't know the best way to run locally Monaco Editor Website but I used below config.

```json
{
    "loaderUrl": "http://localhost:8080/node_modules/monaco-editor-core/dev/vs/loader.js",
    "loaderConfigPaths": {
        "vs": "http://localhost:8080/node_modules/monaco-editor-core/dev/vs",
        "vs/fillers/monaco-editor-core": "http://localhost:8080/out/languages/amd-tsc/fillers/monaco-editor-core-amd",
        "vs/language": "http://localhost:8080/out/languages/amd-tsc/language",
        "vs/basic-language": "http://localhost:8080/out/languages/amd-tsc/basic-language"
    },
    "codiconUrl": "http://localhost:8080/node_modules/monaco-editor-core/dev/vs/base/browser/ui/codicons/codicon/codicon.ttf",
    "monacoTypesUrl": "http://localhost:8080/out/monaco-editor/monaco.d.ts"
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
 	"devDependencies": {
 		"@types/mocha": "^9.1.0",
 		"@types/shelljs": "^0.8.11",
+		"@typescript/ata": "^0.9.3",
 		"@typescript/vfs": "^1.3.5",
 		"chai": "^4.3.6",
 		"clean-css": "^5.2.4",
@@ -73,6 +74,9 @@
 		"vscode-uri": "3.0.3",
 		"webpack": "^5.76.0",
 		"yaserver": "^0.4.0"
+	},
+	"overrides": {
+		"typescript": "^5.0.2"
 	},
 	"alias": {
 		"process": false,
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -74,6 +75,9 @@
 		"webpack": "^5.76.0",
 		"yaserver": "^0.4.0"
 	},
+	"overrides": {
```
tamayika, line 78: `@typescript/ata` currently requires typescript 4.x as peerDependency.
If this PR is going to accepted and merged, I will make a PR to `@typescript/ata` to relax requirements.

hediet:
Thanks for the PR! I'm not convinced though this should be part of the default monaco editor.

tamayika:
I believe that having this feature in the Monaco Editor is crucial when writing JavaScript/TypeScript for CommonJS.
Without this feature, the user would need to have a TypeScript dependency in order to implement it.
This would result in a larger bundled JavaScript file size due to the additional TypeScript code.

PR #3956, opened 2023-05-11
Expose JSON worker

### LINKED ISSUE #727: Expose access to JSON worker
`monaco.languages.typescript` has methods for getting the JS and TS workers, but there is no analogous method for JSON. This seems to imply that getting access to the diagnostic messages requires mining the decorator information or running a second language server.

DavidMulder0:
Fix #727

Exposes the JSON worker and the parseDocument function from the JSON Worker.

This is useful when editors want to use the JSON Worker's ability to parse the document, as using a separate parser can result in inconsistencies and does double the work.

(I had a **lot** of trouble getting this repository to build properly. Best I can tell the instructions in CONTRIBUTING should be significantly updated, and certain things just don't seem to work (the standard playground linked execution seems to never work with the JSON worker)).


DavidMulder0:

DavidMulder0:
Worth noting, this mirrors exactly what the typescript/javascript worker do and how they expose their worker.

DavidMulder0:
And as an example of what this allows

```typescript
monaco.languages.registerCodeLensProvider('json', {
    provideCodeLenses: async function (model, token) {
      const getWorker = await monaco.languages.json.getWorker();
      const worker = await getWorker(model.uri.toString());
      const schema = await worker.getMatchingSchemas(model.uri.toString());
      return {
        lenses: schema
          .filter((item) => item.schema.format === 'ipv4')
          .map((schemaItem) => {
            const start = model.getPositionAt(schemaItem.node.parent.offset);
            const end = model.getPositionAt(
              schemaItem.node.parent.offset + schemaItem.node.parent.length
            );
            return {
              range: {
                startLineNumber: start.lineNumber,
                startColumn: start.column,
                endLineNumber: end.lineNumber,
                endColumn: end.column,
              },
              id: 'First Line',
              command: {
                id: commandId,
                title: 'Select VM',
              },
            };
          }),
        dispose: () => {},
      };
    },
    resolveCodeLens: function (model, codeLens, token) {
      return codeLens;
    },
});
```

hediet:
> (I had a lot of trouble getting this repository to build properly. Best I can tell the instructions in CONTRIBUTING should be significantly updated, and certain things just don't seem to work (the standard playground linked execution seems to never work with the JSON worker)).

Sorry to hear that. What problems did you run into?

DG4ever:
Any news on this pull request? This is exactly what I am looking for.

src/language/json/monaco.contribution.ts
```diff
@@ -4,7 +4,8 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as mode from './jsonMode';
-import { Emitter, IEvent, languages } from '../../fillers/monaco-editor-core';
+import { Emitter, IEvent, languages, Uri } from '../../fillers/monaco-editor-core';
+import type { JSONWorker } from './jsonWorker';
```
aeschli, line 8: We want that all exported types are in monaco.contribution.ts. So also a interface describing the  `JSONWorker` should be defined here.
Can you add a IJSONWoker. I suggest it is limited to the function you want to expose.

aeschli (changes requested):

Issue #3151, opened 2022-06-17
[Bug] typescript worker errors when using `customWorkerPath` in module workers


okikio:
### Reproducible in vscode.dev or in VS Code Desktop?

- [X] Not reproducible in [vscode.dev](https://vscode.dev) or VS Code Desktop

### Reproducible in the monaco editor playground?

- [X] Not reproducible in [the monaco editor playground](https://microsoft.github.io/monaco-editor/playground.html)

### Actual Behavior

The typescript worker crashes when using `customWorkerPath`

```ts
monaco.languages.typescript.typescriptDefaults.setWorkerOptions({
  customWorkerPath: "/ts-worker-factory.js"
});
```
due to `importScript` not being allowed in module workers. A possible solution would be to use the `TypescriptWorker` class to manually extend the typescript language worker, but as monaco doesn't export `TypescriptWorker` there is nothing I can do.

Right now monaco only exports the `create` method from the [...esm/vs/language/typescript/ts.worker.js](https://unpkg.com/browse/monaco-editor@0.33.0/esm/vs/language/typescript/ts.worker.js), which supports `customWorkerPath` via `importScripts`. 

The problem is that for module workers `importScripts` isn't supported, so all you get is an error if you try to use `customWorkerPath` in module workers.

### Expected Behavior

To solve this issue, monaco should export the `TypescriptWorker` class (together with any necessary dependencies) so it can easily be extended by the user, while avoiding the need to call `importScripts` in a module workers.

The source code for [langauage/tsWorker.ts](https://github.com/microsoft/monaco-editor/blob/main/src/language/typescript/tsWorker.ts) shows that the expected behavior is an export, all that would be required would be for the [languages/typescript/ts.worker.ts](https://github.com/microsoft/monaco-editor/blob/main/src/language/typescript/ts.worker.ts) file to add 

```ts
export * from "./tsWorker";
``` 

to the source code.

### Additional Context

This is a problem that affects all the language workers for monaco, it would be best to add exports for all the supported languages in monaco, so extending the source is much easier.

okikio:
I have created a pr that fixes this issue #3152

hediet:
To be honest, I don't fully understand the current worker implementation we have, but I have the feeling it is quite messy.
But I also think that this issue is not about the real problem (which I also don't fully understand), but already about a problem in some solution for the real problem.

[...]

PR #3488, opened 2023-01-05
Export custom TypeScript worker variables

remcohaszing:
A custom TypeScript worker factory function gets passed `TypeScriptWorker`, `ts`, and `libFileMap`. It also exports `create` using ESM. However, this can’t be used, because the custom worker factory is only called if a custom worker path is defined.

Exposing these members, means a worker can be defined using ESM.

A practical use case:

```js
/**
 * @typedef {import('monaco-editor').languages.typescript.CompilerOptions} CompilerOptions
 * @typedef {import('monaco-editor').languages.typescript.IExtraLibs} IExtraLibs
 * @typedef {import('monaco-editor').languages.typescript.InlayHintsOptions} InlayHintsOptions
 * @typedef {import('monaco-editor').languages.typescript.TypeScriptWorker} TypeScriptWorker
 * @typedef {import('monaco-editor').worker.IWorkerContext} IWorkerContext
 * @typedef {import('typescript').LanguageServiceHost} LanguageServiceHost
 *
 * @typedef {object} CreateData
 * @property {CompilerOptions} compilerOptions The TypeScript compiler options configured by the user.
 * @property {string} customWorkerPath The path to a custom worker.
 * @property {IExtraLibs} extraLibs Additional libraries to load.
 * @property {InlayHintsOptions} inlayHintsOptions The TypeScript inlay hints options.
 *
 * @typedef {TypeScriptWorker & LanguageServiceHost} MDXWorker
 * @typedef {new (ctx: IWorkerContext, createData: CreateData) => MDXWorker} TypeScriptWorkerClass
 */

import {createMdxLanguageService} from '@mdx-js/language-service'
// @ts-expect-error This module is untyped.
import {initialize} from 'monaco-editor/esm/vs/editor/editor.worker.js'
// @ts-expect-error This module is untyped.
import {create} from 'monaco-editor/esm/vs/language/typescript/ts.worker.js'

/**
 * @param {TypeScriptWorkerClass} TypeScriptWorker
 * @returns {TypeScriptWorkerClass} A custom TypeScript worker which knows how to handle MDX.
 */
function worker(TypeScriptWorker) {
  return class MDXWorker extends TypeScriptWorker {
    _languageService = createMdxLanguageService(
      // @ts-expect-error This is globally defined in the worker.
      ts,
      this
    )
  }
}

// @ts-expect-error This is missing in the Monaco type definitions.
self.customTSWorkerFactory = worker

// Trick the TypeScript worker into using the `customTSWorkerFactory`
[...]

PR #4035, opened 2023-06-21
Add `customWorkerFactory` to customize tsworker directly

x6eull:
Add a way to customize typescript worker in module workers, avoiding importScripts (which the customWorkerPath uses to extend the typescript worker).

Hoping to provide some progress to https://github.com/microsoft/monaco-editor/issues/3151.

### DIFF
--- a/src/language/typescript/monaco.contribution.ts
+++ b/src/language/typescript/monaco.contribution.ts
@@ -6,6 +6,7 @@
 import type * as mode from './tsMode';
 import { typescriptVersion as tsversion } from './lib/typescriptServicesMetadata'; // do not import the whole typescriptServices here
 import { languages, Emitter, IEvent, IDisposable, Uri } from '../../fillers/monaco-editor-core';
+import { CustomTSWebWorkerFactory } from './tsWorker';
 
 //#region enums copied from typescript to prevent loading the entire typescriptServices ---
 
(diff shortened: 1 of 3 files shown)


LonelyDagger:

hediet:
Afaik this does not work, as `createData` must be serializable. You cannot send over your customization function from the page to the worker.

hediet (approved):
I think it is a little bit ugly that you pass in unserializable data into the create function to override the TypeScriptWorker (and I hope nobody tries to pass in this data from the host side that creates the worker), but I don't see a much nicer solution.

chrmarti (approved):

x6eull:

hediet:
I actually think https://github.com/microsoft/monaco-editor/pull/3488 is a better solution.

x6eull:
I had no idea that `createData` must be serializable when I created the pr. I couldn't build a complete package and test it, so I just followed the step of `customWorkerPath`. If `createData` must be serializable, this approach won't work since I intend to use it in the script that create the worker (as well as the editor).

hediet:
>  If createData must be serializable,

It does not have to be serializable when called from a script that imports the worker, but the data the host passes in has to be serializable. So it would have worked in your use-case, but it could confuse users.
https://github.com/microsoft/monaco-editor/pull/3488 just exports everything needed for a solution that does not cause this confusion.

Thanks for coming up with this PR anyway! :)

PR #4037, opened 2023-06-21
Fix lorem ipsum text on marketing page.

steventrouble:
The github.io page for this has lorem ipsum text in a prominent position on the page. This replaces it with real text.


steventrouble:

hediet:
Thanks! Interesting how nobody noticed this...

hediet (approved):

connor4312 (approved):

PR #4087, opened 2023-07-18
Document some basic concepts

### LINKED ISSUE #4084: [Feature Request] Explain basic concepts
### Context

- [X] This issue is not a bug report. *(please use a different template for reporting a bug)*
- [X] This issue is not a duplicate of an existing issue. *(please use the [search](https://github.com/microsoft/monaco-editor/issues) to find existing issues)*

### Description

On my repositories that contain Monaco editor, I sometimes get questions about Monaco editor from users who don’t understand some of the basic concepts. People think of it as a code editor, but really to use Monaco editor, people must have some basic understanding of:

- URIs
- Models
- Editor instances
- Providers
- Disposables

A little write-up on this would be nice. I could provide one if you want. Just let me know if this belongs on the wiki or readme.

_No response_

_No response_

remcohaszing:
I think this captures the basics. I’m open to suggestions.

IMO this is brief enough to keep in the readme, but I can move it into `/docs/basic-concepts` or so and link that if that’s preferable.

Closes #4084

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,32 @@ You will get:
 
 It is recommended to develop against the `dev` version, and in production to use the `min` version.
 
+## Concepts
+
+Monaco editor is best known for being the text editor that powers VS Code. However, it’s a bit more nuanced. Some basic understanding about the underlying concepts is needed to use Monaco editor effectively.
+
+### Models
+
+Models are at the heart of Monaco editor. It’s what you interact with when managing content. A model represents a file that has been opened. This could represent a file that exists on a file system, but it doesn’t have to. For example, the model holds the text content, determines the language of the content, and tracks the edit history of the content.
+
+### URIs
+
+Each model is identified by a URI. This is why it’s not possible for two models to have the same URI. Ideally when you represent content in Monaco editor, you should think of a virtual file system that matches the files your users are editing. For example, you could use `file:///` as a base path. If a model is created without a URI, it’s URI will be `inmemory://model/1`. The number increases as more models are created.
+
+### Editors
+
+An editor is a user facing view of the model. This is what gets attached to the DOM and what your users see visually. Typical editor operations are displaying a model, managing the view state, or executing actions or commands.
+
+### Providers
+
+Providers provide smart editor features, also known as IntelliSense. For example, this includes completion and hover information. It is not the same as, but often maps to [language server protocol](https://microsoft.github.io/language-server-protocol) features.
+
+Providers work on models. Some IntelliSense depends on the file URI. For example, for TypeScript to resolve imports, or for JSON IntelliSense to determine which JSON schema to apply to which model. So it’s important to choose proper model URIs.
+
+### Disposables
+
+Many Monaco related objects often implement the `.dispose()` method. This method is intended to perform cleanups when a resource is no longer needed. For example, calling `model.dispose()` will unregister it, freeing up the URI for a new model. And editors are typically disposed when changing to another view.
+
 ## Documentation
 
 - Learn how to integrate the editor with these [complete samples](./samples/).
--- a/README.md
+++ b/README.md
@@ -33,15 +33,15 @@ It is recommended to develop against the `dev` version, and in production to use
 
 ## Concepts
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -31,6 +31,32 @@ You will get:
...
+
+Each model is identified by a URI. This is why it’s not possible for two models to have the same URI. Ideally when you represent content in Monaco editor, you should think of a virtual file system that matches the files your users are editing. For example, you could use `file:///` as a base path. If a model is created without a URI, it’s URI will be `inmemory://model/1`. The number increases as more models are created.
+
+### Editors
+
+An editor is a user facing view of the model. This is what gets attached to the DOM and what your users see visually. Typical editor operations are displaying a model, managing the view state, or executing actions or commands.
+
+### Providers
+
+Providers provide smart editor features, also known as IntelliSense. For example, this includes completion and hover information. It is not the same as, but often maps to [language server protocol](https://microsoft.github.io/language-server-protocol) features.
```
hediet, line 52: `IntelliSense` only refers to completions

README.md
```diff
@@ -31,6 +31,32 @@ You will get:
...
+
+Providers work on models. Some IntelliSense depends on the file URI. For example, for TypeScript to resolve imports, or for JSON IntelliSense to determine which JSON schema to apply to which model. So it’s important to choose proper model URIs.
```
hediet, line 54: I'd use "auto-complete", afaik IntelliSense is a Visual Studio only term

README.md
```diff
@@ -31,6 +31,32 @@ You will get:
...
+
+Monaco editor is best known for being the text editor that powers VS Code. However, it’s a bit more nuanced. Some basic understanding about the underlying concepts is needed to use Monaco editor effectively.
+
+### Models
+
+Models are at the heart of Monaco editor. It’s what you interact with when managing content. A model represents a file that has been opened. This could represent a file that exists on a file system, but it doesn’t have to. For example, the model holds the text content, determines the language of the content, and tracks the edit history of the content.
+
+### URIs
+
+Each model is identified by a URI. This is why it’s not possible for two models to have the same URI. Ideally when you represent content in Monaco editor, you should think of a virtual file system that matches the files your users are editing. For example, you could use `file:///` as a base path. If a model is created without a URI, it’s URI will be `inmemory://model/1`. The number increases as more models are created.
```
hediet, line 44:
```suggestion
Each model is identified by a URI. This is why it’s not possible for two models to have the same URI. Ideally when you represent content in Monaco editor, you should think of a virtual file system that matches the files your users are editing. For example, you could use `file:///` as a base path. If a model is created without a URI, its URI will be `inmemory://model/1`. The number increases as more models are created.
```

In general I think we use `'` instead of `’`. I'd keep that for consistency reasons.

README.md
```diff
@@ -31,6 +31,32 @@ You will get:
...
+
+### Providers
+
+Providers provide smart editor features, also known as IntelliSense. For example, this includes completion and hover information. It is not the same as, but often maps to [language server protocol](https://microsoft.github.io/language-server-protocol) features.
+
+Providers work on models. Some IntelliSense depends on the file URI. For example, for TypeScript to resolve imports, or for JSON IntelliSense to determine which JSON schema to apply to which model. So it’s important to choose proper model URIs.
+
+### Disposables
+
+Many Monaco related objects often implement the `.dispose()` method. This method is intended to perform cleanups when a resource is no longer needed. For example, calling `model.dispose()` will unregister it, freeing up the URI for a new model. And editors are typically disposed when changing to another view.
```
hediet, line 58:
```suggestion
Many Monaco related objects often implement the `.dispose()` method. This method is intended to perform cleanups when a resource is no longer needed. For example, calling `model.dispose()` will unregister it, freeing up the URI for a new model. Editors should be disposed to free up resources and remove their model listeners.
```

hediet (changes requested):
Thanks! I left some comments.

README.md
remcohaszing: I thought it applied to Monaco editor too, as it was part of VS Code. Good to know this is not the case. :)

I changed it to _editors features_, as it’s not just auto-complete.

hediet (approved):

hediet:
Thanks!

roblourens (approved):

PR #4151, opened 2023-08-31
fix: editor should not setModelMarkers when allDiagnostics is empty

qingniaotonghua:


qingniaotonghua:
When set setDiagnosticsOptions with {
      noSemanticValidation: true,
      noSyntaxValidation: true,
      noSuggestionDiagnostics: true,
    } to disable default diagnostics,  but  my custom modelMarkers will be cleaned by default diagnostics

PR #4152, opened 2023-08-31
Fix a path and bump editor used for developing website

alexdima:


hediet:
The website CI task automatically installs the latest monaco-editor:
```
  - name: Install most recent version of monaco-editor
    working-directory: website
    run: yarn add monaco-editor
```

But I think it is good to update the version in package.json from time to time.

hediet (approved):

PR #4173, opened 2023-09-11
unwrap qw in wgsl grammar

munrocket:


munrocket:

hediet:
Thanks for this PR! However, I don't see the value of this change, especially as it might discourage the original author of the 
community-contributed grammar to update the keywords (which are probably just copy&pasted from somewhere).
I hope you understand!

PR #4221, opened 2023-10-22
Improve Rust type and keyword detection

LingMan:
See the individual commits for details but in short:
- Add missing Rust types `u128`, `i128`, `usize`, `isize`, and `Result`.
- Add missing Rust keywords `macro_rules` and `'static`.
- Add missing basic macros directly exported from the standard library.
- Drop keywords, types, and constants which are neither part of the core language nor the standard library.


LingMan:
Oh, I didn't realize this repo had a CLA requirement...

PR #4288, opened 2023-12-02
Remove explicit consideration as using|alignas as keywords within a C++ annotation

OfekShilon:
`using` indeed has syntactic meaning within a c++ attribute, but so do `likely`, `deprecated`, `always_inline` and countless others. 
This deletion makes the behavior consistent, and avoids bugs like #3395.


hediet:
The unit tests are failing, can you look into them?

OfekShilon:
> The unit tests are failing, can you look into them?

Yeah sorry, I'll close this PR and suggest a different one some time soon.

PR #4299, opened 2023-12-13
Expose JSON worker

mevisioam:
This PR fixes microsoft#727 and addresses review comments (I think so) from an original PR microsoft#3956 which became stale for some reason.

I just want to get things JSON worker faster to Monaco, because it's needed for my project.

My use case is to provide insights to the end user after analyzing JSON schemas of the edited document. Plus, include the motivation from an original PR: 
> Exposes the JSON worker and the parseDocument function from the JSON Worker.
>
> This is useful when editors want to use the JSON Worker's ability to parse the document, as using a separate parser can result in inconsistencies and does double the work.

### DIFF
--- a/src/language/json/monaco.contribution.ts
+++ b/src/language/json/monaco.contribution.ts
@@ -4,7 +4,8 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as mode from './jsonMode';
-import { Emitter, IEvent, languages } from '../../fillers/monaco-editor-core';
+import { Emitter, IEvent, languages, Uri } from '../../fillers/monaco-editor-core';
+import type { JSONWorker } from './jsonWorker';
 
 // --- JSON configuration and defaults ---------
 
@@ -197,8 +198,12 @@ export const jsonDefaults: LanguageServiceDefaults = new LanguageServiceDefaults
 	modeConfigurationDefault
 );
 
+export const getWorker = (): Promise<(...uris: Uri[]) => Promise<JSONWorker>> => {
+	return getMode().then((mode) => mode.getWorker());
+};
+
 // export to the global based API
-(<any>languages).json = { jsonDefaults };
+(<any>languages).json = { jsonDefaults, getWorker };
 
 // --- Registration to monaco editor ---
 
--- a/src/language/json/monaco.contribution.ts
+++ b/src/language/json/monaco.contribution.ts
@@ -7,69 +7,6 @@ import * as mode from './jsonMode';
 import { Emitter, IEvent, languages, Uri } from 'monaco-editor-core';
 
 // ---- JSON service types ----
-/**
- * Defines a decimal number. Since decimal numbers are very
- * rare in the language server specification we denote the
- * exact range with every decimal using the mathematics
- * interval notations (e.g. [0, 1] denotes all decimals d with
- * 0 <= d <= 1.
- */
-export type decimal = number;
-
-/**
- * Defines an unsigned integer in the range of 0 to 2^31 - 1.
- */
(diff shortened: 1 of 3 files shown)


mevisioam:

mevisioam:
@hediet @aeschli can you take a look? Since you looked at #3956 previously :)

src/language/json/monaco.contribution.ts
```diff
@@ -4,7 +4,8 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as mode from './jsonMode';
-import { Emitter, IEvent, languages } from '../../fillers/monaco-editor-core';
+import { Emitter, IEvent, languages, Uri } from 'monaco-editor-core';
+import * as jsonService from 'vscode-json-languageservice';
```
aeschli, line 8: The types that are used from `jsonService` should be copied to `json/monaco.contribution.ts` so that the d.ts file is self-contained.

aeschli (changes requested):

src/language/json/monaco.contribution.ts
mevisioam: Got it, lemme fix it.

mevisioam: Though, I wonder then why we have pretty much the same setup in https://github.com/microsoft/monaco-editor/blob/main/src/language/typescript/monaco.contribution.ts#L6 :)

mevisioam: Ah, sorry, it's `jsonMode` not `jsonService`.

mevisioam: @aeschli looks crazy, but I'm done, sir :)

aeschli:
Yes, that's a lot of types, I don't like it either. Especially as these types duplicate some of the Monaco types (e.g. IPosition, IRange), but Monaco types 1 based  (the first column and line is 1, not 0), while the LSP types are 0 based.

Not sure what approach to choose:
- Keep the API minimal. What API from the worker do you really need? 
- In typesscript/monaco.contribution.ts the APis just return any. We could do that.

@alexdima What do yo think?

mevisioam:
@aeschli:
> * Keep the API minimal. What API from the worker do you really need?

Think, I could leave it to only `getMatchingSchemas` and `parseDocument` as in the initial PR did, afair.

mevisioam:
@aeschli cleanup a lil bit

src/language/json/monaco.contribution.ts
```diff
@@ -4,8 +4,221 @@
...
+/**
+ * A tagging type for string properties that are actually document URIs.
+ */
+export type DocumentUri = string;
+
+/**
+ * Represents a location inside a resource, such as a line
+ * inside a text file.
+ */
+export interface Location {
```
aeschli, line 218: I think we can remove Location, Range and Position

aeschli (changes requested):

src/language/json/monaco.contribution.ts
mevisioam: @aeschli And more than that. Done.

mevisioam:
@aeschli @alexdima this state shud be fine now.

aeschli:
Looks good!

aeschli (approved):

lramos15 (approved):

mevisioam:
@aeschli thx for the review, when shall this end up in `npm install monaco-editor`? :)

aeschli:
I think @hediet makes a Monaco editor release in the same week VS Code is released. So this should be in the last week of January.

PR #4445, opened 2024-03-27
Update Parcel integration docs

victoralvelais:
The Parcel documentation was out of date with the current Browser ESM Parcel sample and added a lot more complexity to the integration process.


victoralvelais:
@microsoft-github-policy-service agree company="Victory Bytes"

PR #4482, opened 2024-04-25
Call clearFiles on internal EmitOutput diagnostics

jakebailey:
Emitting produces diagnostics. Unfortunately, these are marked as internal so were missed (https://github.com/microsoft/TypeScript/pull/58317)

This change enables the playground to handle emit errors without crashing.


hediet (approved):

hediet:
Thanks!

jakebailey:
The next version of TS will have this be formally public, so some of the cruft will be able to go away when that upgrade happens (but this PR works for older versions too)

jrieken (approved):

PR #4585, opened 2024-07-01
C# raw string literal

jairbubbles:
- Support C# raw string literals (cf https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/raw-string)
- It starts by a minimum of three `"` and end by the same numberf of `#`


jairbubbles:

vgallegob:
I would like to see this one added, it's a much better way to do literals.

PR #4652, opened 2024-08-26
Removing `editor.main.nls.js`

aiday-mar:
Removing `editor.main.nls.js`


hediet (approved):

hediet:
This file is no longer needed, as there is no nls plugin anymore. When the language is English (default language), no additional files need to be loaded.

PR #4664, opened 2024-08-29
Defines setInterval/setTimeout on faked globalThis in unit test

hediet:


joyceerhl (dismissed):

Tyriar (commented):
Is this debt that should go in for September? The less changes after testing the better

Tyriar:
Oh actually this is monaco editor so I don't think it impacts vscode's endgame 👌

aiday-mar:
We decided to undo the package upgrade change here and just push the code change first. 
Then we will publish a new monaco-editor-core package with the fix for the webpack smoke test
Then we will bump the version of the package on main on monaco-editor

aeschli (approved):

PR #4733, opened 2024-10-30
Fix yaml keywords of NULL

weareoutman:
There is two 'Null's in yaml keywords, while one should be 'NULL'.


18568803179:
Hey, I'd love to know how you configure the syntax hints and validations for yaml in monaco. I configured only highlighting... @weareoutman

PR #4751, opened 2024-11-14
docs: change `var require` to `window.require`

Blackman99:
The browser script edtior sample use `var require = ...`

In some senario var would be auto replaced to `const` (ESLint eg.). And would cause worker not found, so as other resource files.

### DIFF
--- a/samples/browser-script-editor/index.html
+++ b/samples/browser-script-editor/index.html
@@ -13,7 +13,7 @@
 		<div id="container" style="width: 800px; height: 600px; border: 1px solid grey"></div>
 
 		<script>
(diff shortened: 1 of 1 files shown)


femboypig:
"In some" but not all of them.

callmegautam:
Good fix. Changing `var require` to `window.require` ensures that the `require` configuration is globally available for Monaco's AMD loader (`loader.js`). This avoids conflicts with local `require` definitions and aligns with how Monaco expects the global `require` to be defined.

callmegautam (approved):
Looks good! Using `window.require` makes it globally accessible – avoids reference errors in some browser environments. Nice improvement!

PR #4938, opened 2025-07-10
Fix memory leaks in webpack plugin from repeated worker compilation

dgaponov:
## Problem

In development mode with Hot Module Replacement (HMR) enabled, workers were being recompiled on every change, leading to memory leaks. This issue was particularly problematic in rspack environments where repeated compilation cycles would cause significant memory accumulation.

https://github.com/web-infra-dev/rspack/issues/8976#issuecomment-2768527126

## Root Cause

The `AddWorkerEntryPointPlugin` was creating new child compilers for worker entry points on every compilation cycle without tracking whether the worker had already been compiled. This resulted in:
- Unnecessary repeated worker compilation
- Memory leaks from accumulated compilation artifacts
- Degraded development experience due to memory pressure

## Solution

- Added `alreadyCompiled` flag to track compilation state
- Prevents worker recompilation after the first successful build
- Maintains existing functionality while eliminating memory leaks

## Changes

- Modified `getCompilerHook` function to include compilation tracking
- Added early return when worker is already compiled
- No breaking changes to existing API

## Testing

- ✅ Verified worker compilation works correctly on first build
- ✅ Confirmed no recompilation on subsequent HMR cycles
- ✅ Tested memory usage remains stable during development
- ✅ Validated compatibility with both webpack and rspack

## Impact

- **Memory Usage**: Significantly reduced memory consumption in dev mode
- **Performance**: Faster HMR cycles by skipping unnecessary worker compilation
- **Developer Experience**: More stable development environment
- **Compatibility**: Works with both webpack and rspack bundlers

Fixes memory leak issues reported in development environments using HMR with worker entry points.


dgaponov:

dgaponov:
@hediet @alexr00
Hi! I'd really appreciate it if you could review my PR. This fix is quite important for my team."

dgaponov:
@hediet @alexr00
UP

dgaponov:
@hediet @alexr00
UP

dgaponov:
[Henning Dieterichs (@hediet)](https://github.com/hediet) [Alex Ross (@alexr00)](https://github.com/alexr00)
UP

PR #4973, opened 2025-09-04
Fix Kotlin number literals

flofriday:
There were a couple of problems in Kotlins number literal parsing.
This PR adapts it to conform to the grammar described in the [Language Spec](https://kotlinlang.org/spec/pdf/kotlin-spec.pdf) chapter _1.2.3 Literals_.

### Unsigned suffixes

In Kotlin 1.5 unsigned numbers were introduced, including the suffixes `u` and `U` for number literals and while the following examples are valid Kotlin now they weren't correctly highlighted:

```kotlin
val u1 = 123u
val u2 = 123U
val u3 = 123uL
val u4 = 0b0101UL
val u5 = 0Xff0088u
```

### Leading dot floating point literals

In Kotlin number literals can start with a leading dot.

```kotlin
val f1 = .1
val f2 = .123
val f3 = .123e+12f
```

### Long suffix must be uppercase
Unlike Java only the uppercase `L` is a valid suffix.

```kotlin
// Valid
val correct = 123L

// Error:  Use 'L' instead of 'l'.
val wrong = 123l
```

### Double suffix doesn't exist

Unlike Java there isn't a suffix to indicate that a number is a double.

```kotlin
// Valid
val correct = 123L

// Error
val wrong1 = 123D
val wrong1 = 123d
```

### DIFF
--- a/src/basic-languages/kotlin/kotlin.test.ts
+++ b/src/basic-languages/kotlin/kotlin.test.ts
@@ -279,6 +279,13 @@ testTokenization('kotlin', [
 
 	[
 		{
+			line: '.123',
+			tokens: [{ startIndex: 0, type: 'number.float.kt' }]
+		}
+	],
+
+	[
+		{
 			line: '0x',
 			tokens: [
 				{ startIndex: 0, type: 'number.kt' },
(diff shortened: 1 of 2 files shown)


flofriday:

hediet (approved):

PR #4974, opened 2025-09-05
Fix Microsoft logo on footer

bit-koda:
Corrected the image path to load the Microsoft logo from the local assets folder, fixing the 404 error on the footer logo display.


hediet:
Thanks! I removed the logo in https://github.com/microsoft/monaco-editor/pull/4996, to avoid problems with forks of this repository.

PR #5030, opened 2025-10-08
Update integrate-esm.md

MaxenceMouchard:
This update fixes the Vite integration note because `self.MonacoEnvironment.getWorkerUrl` is undefined.


MaxenceMouchard:

PR #5062, opened 2025-10-19
Add createModular function to support modular TypeScript worker creation

cp89cyber:
fixes issue#5061


cp89cyber:

hediet:
Unfortunately, it is too difficult for us to review this PR given its size.

PR #5115, opened 2025-11-18
Installs playwright deps via install-deps script

hediet:


hediet:
Actually, this shouldn't be needed.

PR #5136, opened 2025-12-05
Fix a11y color contrast bug in sample

hawkticehurst:
Resolves https://github.com/microsoft/monaco-editor/issues/5074

Updates Colors and Tokens sample to use a comment color (from yellow to dark pink) that meets a11y color contrast requirements.


meganrogge (approved):
Thanks!

Issue #5148, opened 2025-12-16
Remove worker namespace from monaco-editor-core


hediet:
Explore if we can deprecate the monaco editor worker feature and offer other ways (e.g. a library to synchronize models across some transport channel).

Would also fix https://github.com/microsoft/monaco-editor/issues/5049.

PR #5156, opened 2025-12-19
worker-based LSP server

hediet:
Prepares for https://github.com/microsoft/monaco-editor/issues/5148

### DIFF
--- a/monaco-lsp-client/generator/index.ts
+++ b/monaco-lsp-client/generator/index.ts
@@ -1,5 +1,7 @@
 import * as fs from 'fs';
 import * as path from 'path';
+import yargs from 'yargs';
+import { hideBin } from 'yargs/helpers';
 
 /**
  * Utility class for writing formatted code with proper indentation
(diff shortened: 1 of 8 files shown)


giautm:
Hey @hediet, thank you for working on this. 🙏 .

We were using monaco-editor for a long time, but always struggle with the LSP support.

I'm so happy to see this implement in the monaco project. I hope you can finish it soon, and expose enought APIs for public usage.

My situation is about LSP-Server in Webworker (WASM-base), connect with the monaco like you did there.

hediet:
@giautm LSP client support is already in! This is about making it easy to create LSP servers in a webworker.

giautm:
> @giautm LSP client support is already in! This is about making it easy to create LSP servers in a webworker.

Yes, I found the playground after comment there. And I was successful to integrate the LSP client with my LSP Server running in WASM. However, I need some "hack" to inject extra features to the client. I need access to the `_connection` which is not exported right now. Also the connection type is not exposed yet.

```ts
// LSP Client setup
class WasmFeature {
  constructor(c: any) {
    c.capabilities.addStaticClientCapabilities({
      experimental: {
        noFilesystem: true,
      },
    });
  }
  dispose() {}
}

class AtlasLspClient extends lsp.MonacoLspClient {
  createFeatures() {
    const store = super.createFeatures();
    //@ts-ignore
    store.add(new WasmFeature(this._connection));
    //@ts-ignore
    console.log("LSP Features initialized:", store, this._connection);
    return store;
  }
}
```

Issue #3233, opened 2022-08-06
[Feature Request] Add LaTeX and BibTeX language definitions from MSR's https://github.com/koka-lang/madoko


vadimkantorov:
### Context

- [X] This issue is not a bug report. *(please use a different template for reporting a bug)*
- [X] This issue is not a duplicate of an existing issue. *(please use the [search](https://github.com/microsoft/monaco-editor/issues) to find existing issues)*

### Description

@daanx developed https://github.com/koka-lang/madoko/blob/master/styles/lang/latex.json and https://github.com/koka-lang/madoko/blob/master/styles/lang/bibtex.json . It would be nice to upstream them into this repo

Originally posted at: https://github.com/koka-lang/madoko/issues/71

josealvarez97:
This is a good idea! Bumping this up.

pranftw:
Any update on when this would be available?

sitanshu05:
Any updates on this??

PR #5271, opened 2026-03-27
add latex and mathjax languages including samples

boombard:
Partially addresses #3233 (no bibtex)

Used the following sources:
* Latex: https://github.com/koka-lang/madoko/blob/master/styles/lang/latex.json
* Mathjax: https://www.cmor-faculty.rice.edu/~heinken/latex/symbols.pdf (amongst others).


boombard:

PR #5272, opened 2026-03-28
Fix Python multi-line f-strings.

### LINKED ISSUE #4601: [Bug] new support for Python f-strings breaks with multi-line f-strings

- [X] Not reproducible in [vscode.dev](https://vscode.dev) or VS Code Desktop

https://microsoft.github.io/monaco-editor/playground.html?source=v0.50.0#XQAAAAIRAgAAAAAAAABBqQkHQ5NjdMjwa-jY7SIQ9S7DNlzs5W-mwj0fe1ZCDRFc9ws9XQE0SJE1jc2VKxhaLFIw9vEWSxW3yscw0dONKAEiAKctgAboa7AaZZDHLgrLqPRv_fRv2RyIokydlTdTYccsu6v6jDrc8jQ8urtpKMM2vyRZMFv2XBDZhYVK59hZxbzQYIV5CtdMNnrXBtIuhs4pit-oL9QyaQILWijHfHxfkUWefrcArGLpBhdoYjkZiGMY86lvdZUBvKZdsphejnt8Nh47ZzyPeslH-Tbe20EaxqbYdwnBi7VKWQHQ2xDAOVNcX6EW8BTUElF6RN3c9dKt_ztqq1rNEhGHDgAU5zXWFKRChN1RN_VvIEhPxjjyr6w2-1TXQgES6nZ2R8rHaswkY1d1ECmaIHNzL-qG-TR6GJCHFt4qvP0wOFVlbInDK8MY99y7BjDnexwGkeZL10eyfHBdbn79Cd76_8_NKHVLANCFh4N0snUs9HVXdkXBho1Ht8LOMLKR4L9YkIRlpqi80DoC7_q4yNg

```typescript
single_line_f_string f'''nothing {'to see'} here'''

multi_line_f_string = f"""first line looks fine
{'uh oh'}
now the highlighting is broken down here :(
"""

also = "it's broken highlighting for everything after"
four = 2 + 2
```

### Reproduction Steps

1. Configure an editor using the built-in python language
2. Enter a multi-line f-string which has an expression after the 1st line

The portion of the f-string after the expression is highlighted as python code instead of a string, and then the closing f-string delimiter causes code after that to be highlighted as if it were a string

### Expected Behavior

_No response_

### Additional Context

CleanShot 2024-07-12 at 11 08 38@2x

yilei:
This routes `f'''` and `f"""` to dedicates states, which will be closed on matching `'''` and `"""` quotes. It also adds the supported raw-string variant prefixes like `rf` / `rF` / ...

This should fix #4601.

### DIFF
--- a/src/languages/definitions/python/python.test.ts
+++ b/src/languages/definitions/python/python.test.ts
@@ -253,5 +253,130 @@ testTokenization('python', [
 				{ startIndex: 6, type: 'string.escape.python' }
 			]
 		}
+	],
+
+	// https://github.com/microsoft/monaco-editor/issues/4601
+	// Multiline triple-quoted f-strings.
+	[
+		{
+			line: 'x = f"""first line',
+			tokens: [
+				{ startIndex: 0, type: 'identifier.python' },
+				{ startIndex: 1, type: 'white.python' },
+				{ startIndex: 2, type: '' },
+				{ startIndex: 3, type: 'white.python' },
+				{ startIndex: 4, type: 'string.escape.python' },
+				{ startIndex: 8, type: 'string.python' }
+			]
+		},
+		{
+			line: 'still {var} string',
+			tokens: [
+				{ startIndex: 0, type: 'string.python' },
+				{ startIndex: 6, type: 'identifier.python' },
+				{ startIndex: 11, type: 'string.python' }
+			]
+		},
+		{
+			line: 'last line"""',
+			tokens: [
+				{ startIndex: 0, type: 'string.python' },
+				{ startIndex: 9, type: 'string.escape.python' }
+			]
+		},
+		{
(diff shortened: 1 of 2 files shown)


nicolasakf:
Hi can we please merge this PR. I patched it locally and it solved the problem for me.

rdejongeviktor:
Any updates on this?

jonathanKingston:
I also hit this and independently reached the same fix as — routing f'''/f""" into dedicated multi-line states — which I now ship as a local grammar override, verified against real tokenization output. One gap worth fixing while you're in there: {{/}} literal braces match the interpolation rule and, in a multi-line state, swallow every following line until a }. Adding [/\{\{|\}\}/, 'string'] before the interpolation rule fixes it.

PR #5303, opened 2026-04-18
add basic syntax support for cedar

Swolebrain:
This translates the TextMate grammars used in the official Cedar vscode plugin to monarch, adding support for the cedar policy language and cedar schemas.


Swolebrain:

PR #5341, opened 2026-05-25
fix: Populate diagnostic fields on send codeAction

### LINKED ISSUE #5342: [Bug] [LSP] `textDocument/codeAction` request omits diagnostic fields the server needs to produce quickfixes (`data`, `code`, `source`, …)

- [x] Not reproducible in [vscode.dev](https://vscode.dev) or VS Code Desktop

- [x] Not reproducible in [the monaco editor playground](https://microsoft.github.io/monaco-editor/playground.html)

### Summary

`LspCodeActionProvider.provideCodeActions` constructs `context.diagnostics` for the outgoing `textDocument/codeAction` request from `context.markers`, but forwards only `range`, `message`, and `severity`. LSP diagnostic fields that servers commonly key off — `code`, `source`, `tags`, `relatedInformation`, `codeDescription`, and especially `data` (the server-roundtripped opaque payload) — are dropped, even though most are present on `IMarker` and the `data` / `codeDescription` payloads are still available on the original `Diagnostic` objects this client just received from `publishDiagnostics` / `textDocument/diagnostic`.

### Effect

For any server that uses `Diagnostic.data` to carry quickfix metadata (the LSP 3.16+ mechanism for [diagnostic-targeted code actions](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeAction), gated on `dataSupport: true` — which this client already advertises in `LspDiagnosticsFeature.ts:25`), the server can't identify the diagnostic in the codeAction request and returns no `quickfix.*` actions. The Monaco lightbulb only shows source-level actions (`source.fixAll.*`, `source.organizeImports.*`), and the user-facing "Remove unused import" / "Add missing return" / etc. quickfixes never appear.

Concrete real-world example: [`ruff server`](https://github.com/astral-sh/ruff) attaches `data` to every diagnostic. Without it ruff returns zero quickfix actions, even though the same server emits all of them when a fully-formed diagnostic is round-tripped back.

### Reproduction (3-shape probe against ruff)

Initialize → `didOpen` (`"import os\n\nprint('hi')\n"` → F401 unused import) → wait for `publishDiagnostics`. Then send three `textDocument/codeAction` requests with different `context.diagnostics`:

| Shape | `context.diagnostics` payload | Result |
|---|---|---|
| **A** — full Diagnostic as ruff sent it | `[{ range, message, severity, code: "F401", source: "Ruff", data: {…} }]` | 4 actions, including 2 `quickfix.*` ("Remove unused import: 'os'", "Disable for this line") |
| **B** — stripped to `range`/`message`/`severity` (what `LspCodeActionProvider` sends today) | `[{ range, message, severity }]` | 2 source-only actions, no quickfixes |
| **C** — empty | `[]` | 2 source-only actions, no quickfixes |

Shape B is what every Monaco user sees today. `range + message` alone is theoretically enough to route the request, but in practice servers like ruff key off `data` and silently fall back to source actions.

### Root cause

[`monaco-lsp-client/src/adapters/languageFeatures/LspCodeActionFeature.ts:80-85`](https://github.com/microsoft/monaco-editor/blob/main/monaco-lsp-client/src/adapters/languageFeatures/LspCodeActionFeature.ts#L80-L85):

```ts
diagnostics: context.markers.map(marker => ({
    range: this._client.bridge.translateRange(model, monaco.Range.lift(marker)),
    message: marker.message,
    severity: toLspDiagnosticSeverity(marker.severity),
})),
```

`code`, `source`, `tags`, and `relatedInformation` are recoverable directly from `marker` — they're already preserved by `toDiagnosticMarker` in [`common.ts:372-399`](https://github.com/microsoft/monaco-editor/blob/main/monaco-lsp-client/src/adapters/languageFeatures/common.ts#L372-L399) when LSP diagnostics are turned into IMarkers. But `data` and `codeDescription` cannot survive the round-trip through `IMarker` because Monaco's marker model doesn't have fields for them. They're available on the original `Diagnostic` objects in `LspDiagnosticsFeature`, but that feature doesn't expose them.

### Proposed fix

1. New `DiagnosticsCache` (URI → `Diagnostic[]`) populated by `LspDiagnosticsFeature` whenever it receives diagnostics (push and pull paths), cleaned on `model.onWillDispose`.
2. Expose it via `LspConnection` so any feature can read it.
3. `LspCodeActionProvider.provideCodeActions` looks up each marker's matching `Diagnostic` (by URI + range + message) and forwards the original object verbatim. Markers without a cache hit (e.g. set by an unrelated owner) fall back to a synthesized Diagnostic that still passes through `code`, `source`, `tags`, and `relatedInformation` from `IMarker` — so foreign markers still contribute the fields they have.

Diff in the linked PR.

### LSP 3.17 spec reference

Per [§ Diagnostic](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic):

> A data entry field that is preserved between a `textDocument/publishDiagnostics` notification and `textDocument/codeAction` request. […] Servers are allowed to attach information to a diagnostic, e.g. for purposes of identifying it later.

The current implementation makes this round-tripping impossible.

```typescript

```

### Reproduction Steps

_No response_

_No response_

### Expected Behavior

_No response_

### Additional Context

_No response_

apatenotre:
# [LSP] Round-trip `Diagnostic.data` and other fields through `textDocument/codeAction`

Fixes #5342.

`LspCodeActionProvider.provideCodeActions` was sending `context.diagnostics` entries containing only `range`, `message`, and `severity` — dropping `code`, `source`, `tags`, `relatedInformation`, `codeDescription`, and `data`. The LSP spec requires `data` to be round-tripped (this client even advertises `dataSupport: true`), and real-world servers (e.g. ruff) silently drop quickfix actions when it's missing.

### Approach

1. **New `DiagnosticsCache`** (`monaco-lsp-client/src/adapters/DiagnosticsCache.ts`) keyed by Monaco URI string.
2. **Populated by `LspDiagnosticsFeature`** from both push (`publishDiagnostics`) and pull (`textDocument/diagnostic`) paths, including the related-documents branch. Entries removed on `model.onWillDisposeModel`.
3. **Exposed on `LspConnection.diagnosticsCache`** so any feature can read it.
4. **`LspCodeActionProvider.provideCodeActions`** looks up each marker's matching cached `Diagnostic` (by range + message in LSP coordinates) and forwards the full object. Markers without a cache hit (set by some other source — e.g. ESLint extension setting markers under a different owner) fall back to a synthesized Diagnostic that still includes `code` / `source` / `tags` / `relatedInformation` from `IMarker`.

Cache keying is by Monaco URI (`model.uri.toString()`) throughout, to keep cleanup trivial via `monaco.editor.onWillDisposeModel` regardless of whether diagnostics came in via push or pull.

### Verification done locally

In a real Next.js + Monaco app using `monaco-editor` `0.55.1`:

1. **Before the fix** — WS frames captured in DevTools show outgoing `textDocument/codeAction.context.diagnostics` entries containing only `range`, `message`, `severity`. Ruff returns 2 source-only actions. Monaco lightbulb shows no quickfixes on the `F401` unused-import diagnostic.
2. **After the fix** (applied via a local transport-proxy workaround that mirrors what this PR does inside the client) — outgoing entries include `code: "F401"`, `source: "Ruff"`, `data: {…}`. Ruff returns 4 actions including 2 `quickfix.*`. "Remove unused import: 'os'" and "Disable for this line" appear in the lightbulb.
3. **Regression check** — basedpyright (which doesn't use `data`) is unchanged; its hover, definition, code actions, and diagnostics all still work.

### Notes

- No test added — `monaco-lsp-client/` currently has no test infrastructure (no test script, no framework dep, no `test/` folder). Happy to add one if you pick a framework; the assertion shape is: stub the LSP transport to capture outgoing `textDocument/codeAction` payloads, push a Diagnostic with `data`/`code`/`source` via `publishDiagnostics`, set a marker matching that range, trigger a code-action request, assert the captured outgoing payload's `context.diagnostics[0]` matches the original verbatim.
- The cache is keyed by `model.uri.toString()` (Monaco URI), not LSP URI, so that cleanup via `monaco.editor.onWillDisposeModel` is a one-liner. Both producers translate to Monaco URI before caching.
- For markers under a different owner (e.g. a separate extension's marker collection), the lookup misses and we fall back to the synthesized Diagnostic — preserving the fields IMarker actually carries. This is strictly more information than the current code sends.
- Independent of #<bug-2-issue-number> (the dispose leak). Either can land first.

### DIFF
--- a/monaco-lsp-client/src/adapters/DiagnosticsCache.ts
+++ b/monaco-lsp-client/src/adapters/DiagnosticsCache.ts
@@ -1,0 +1,32 @@
+import { Diagnostic } from '../../src/types';
+
+/**
+ * Per-document cache of the last set of LSP `Diagnostic` objects received
+ * from the server (via `publishDiagnostics` or `textDocument/diagnostic`).
+ *
+ * Monaco's `IMarker` model can't carry the full Diagnostic shape: `data` and
+ * `codeDescription` have no corresponding IMarker field, so once a Diagnostic
+ * is translated to a marker those values are lost. Features that need to
+ * round-trip the original Diagnostic to the server (notably
+ * `textDocument/codeAction`, which the spec requires to include `data`) read
+ * from this cache to recover the missing fields.
+ *
+ * Keyed by Monaco URI string (`model.uri.toString()`) so cleanup can be done
+ * directly from `monaco.editor.onWillDisposeModel` without needing the
+ * text-model bridge.
+ */
+export class DiagnosticsCache {
+    private readonly _byUri = new Map<string, readonly Diagnostic[]>();
+
+    public set(uri: string, diagnostics: readonly Diagnostic[]): void {
+        this._byUri.set(uri, diagnostics);
+    }
+
+    public get(uri: string): readonly Diagnostic[] | undefined {
+        return this._byUri.get(uri);
+    }
+
+    public delete(uri: string): void {
+        this._byUri.delete(uri);
+    }
+}
--- a/monaco-lsp-client/src/adapters/LspClient.ts
+++ b/monaco-lsp-client/src/adapters/LspClient.ts
@@ -1,90 +1,97 @@
-import { IMessageTransport, TypedChannel } from "@hediet/json-rpc";
-import { LspCompletionFeature } from "./languageFeatures/LspCompletionFeature";
-import { LspHoverFeature } from "./languageFeatures/LspHoverFeature";
-import { LspSignatureHelpFeature } from "./languageFeatures/LspSignatureHelpFeature";
-import { LspDefinitionFeature } from "./languageFeatures/LspDefinitionFeature";
-import { LspDeclarationFeature } from "./languageFeatures/LspDeclarationFeature";
-import { LspTypeDefinitionFeature } from "./languageFeatures/LspTypeDefinitionFeature";
-import { LspImplementationFeature } from "./languageFeatures/LspImplementationFeature";
-import { LspReferencesFeature } from "./languageFeatures/LspReferencesFeature";
-import { LspDocumentHighlightFeature } from "./languageFeatures/LspDocumentHighlightFeature";
-import { LspDocumentSymbolFeature } from "./languageFeatures/LspDocumentSymbolFeature";
-import { LspRenameFeature } from "./languageFeatures/LspRenameFeature";
-import { LspCodeActionFeature } from "./languageFeatures/LspCodeActionFeature";
-import { LspCodeLensFeature } from "./languageFeatures/LspCodeLensFeature";
-import { LspDocumentLinkFeature } from "./languageFeatures/LspDocumentLinkFeature";
-import { LspFormattingFeature } from "./languageFeatures/LspFormattingFeature";
-import { LspRangeFormattingFeature } from "./languageFeatures/LspRangeFormattingFeature";
-import { LspOnTypeFormattingFeature } from "./languageFeatures/LspOnTypeFormattingFeature";
-import { LspFoldingRangeFeature } from "./languageFeatures/LspFoldingRangeFeature";
-import { LspSelectionRangeFeature } from "./languageFeatures/LspSelectionRangeFeature";
-import { LspInlayHintsFeature } from "./languageFeatures/LspInlayHintsFeature";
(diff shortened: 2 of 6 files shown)


apatenotre:

PR #5447, opened 2026-08-10
fix(diff-editor): support originalEditable in inline diff mode (#5302)

### LINKED ISSUE #5302: [Bug] Editing original model in inline diff viewer not possible even if it should be based on configuration options

- [x] Not reproducible in [vscode.dev](https://vscode.dev) or VS Code Desktop

https://microsoft.github.io/monaco-editor/playground.html?source=v0.55.1#XQAAAAJjAgAAAAAAAABBqQkHQ5NjdMjwa-jY7SIQ9S7DNlzs5W-mwj0fe1ZCDRFc9ws9XQE0SJE1jc2VKxhaLFIw9vEWSxW3yscw1ywOc3g3TZokiQe3t547J-9ebB9cYURQV2C_HyzmFKq-zHr7CL5opGlpi_AOvKvzZ3i3ITvDrqy_CgWLe8nzPfRcjGsigwE9_J2qOtPn88pmhu9_yKIbFcbbOldJxbsZ9LkVEh_hzRCzlJWgqN9MOF72YHODV1uZEj0HtZq3JRXxYNUUXVYqGIDguR07WhPWmy4POSvcvO4BR2YcCzfmfK5ZDnxw2Fbf91l4zAyMZHYEiY5QdFX4EoNu8IIpxG26RVpzI0DTeg-y9ZJdmFRQ2WtKP9Z-dTEyLOkyISB0vZeBBTroraw10xURz68D9vhkg3m0pjN7m8l0LPLIm1ulGWNVRVw2of6eexD9nI1XRacX_3t1zgA

```typescript
var originalModel = monaco.editor.createModel(
	"This line is removed on the right.\njust some text\nabcd\nefgh\nSome more text",
	"text/plain"
);

var modifiedModel = monaco.editor.createModel(
	"just some text\nabcz\nzzzzefgh\nSome more text\nThis line is removed on the left.",
	"text/plain"
);

var diffEditor = monaco.editor.createDiffEditor(
	document.getElementById("container"),
	{
		renderSideBySide: false,
		originalEditable: true,
		readOnly: true, // Can be either true or false
	}
);

diffEditor.setModel({
	original: originalModel,
	modified: modifiedModel,
});
```

### Reproduction Steps

With the code above, attempt to edit the original model in the diff editor.

The original model is not editable in an inline diff editor even though `originalEditable` is `true`. If the diff editor wasn't inline, the original model would be editable. 

### Expected Behavior

The original model should be editable.

### Additional Context

_No response_

veeresh4520:
## Description

Fixes #5302.

The original model of an inline diff editor could not be edited even when `originalEditable` was set to `true`.

This change fixes the issue while preserving the existing `readOnly` behavior and adds a regression test to ensure the original model remains editable when the configuration allows it.

## Changes

* Fixed editing of the original model in inline diff editors.
* Added a regression test covering `originalEditable` in inline diff mode.
* Preserved existing behavior for side-by-side diff editors and read-only configurations.

## Testing

* Added a regression test for the reported configuration.
* Ran the relevant diff editor tests.
* Ran the applicable repository validation checks.

### DIFF
--- a/test/diffEditor.test.ts
+++ b/test/diffEditor.test.ts
@@ -1,0 +1,117 @@
+import test from 'node:test';
+import assert from 'node:assert';
+import * as monaco from '../node_modules/monaco-editor-core/esm/vs/editor/editor.api.js';
+
+test('diff editor originalEditable option in inline mode (issue #5302)', async (t) => {
+	await t.test('allows editing original model when originalEditable is true and readOnly is false in inline diff mode', () => {
+		const originalModel = monaco.editor.createModel('original content', 'text/plain');
+		const modifiedModel = monaco.editor.createModel('modified content', 'text/plain');
+		const container = document.createElement('div');
+		document.body.appendChild(container);
+
+		const diffEditor = monaco.editor.createDiffEditor(container, {
+			renderSideBySide: false,
+			originalEditable: true,
+			readOnly: false,
+			renderOverviewRuler: false,
+		});
+
+		diffEditor.setModel({ original: originalModel, modified: modifiedModel });
+
+		const origEditor = diffEditor.getOriginalEditor();
+		const modEditor = diffEditor.getModifiedEditor();
+
+		assert.strictEqual(origEditor.getOption(monaco.editor.EditorOption.readOnly), false);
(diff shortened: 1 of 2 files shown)


veeresh4520:
@microsoft-github-policy-service agre e

